PR #16, opened 2019-02-19
Coloring in dpkg-buildpackage is broken when output in Jenkins

pzakha:
We want to stop coloring the output of `dpkg-buildpackage` when ran from Jenkins as it causes coloring issues when building certain packages (e.g. `bcc`). Another side-effect of coloring issues is that invisible characters can get copied along when copying Jenkins console strings.

## Testing
- http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/pre-push/69
- make check


pzakha:
cc: @jgallag88

Issue #7, opened 2018-12-26
intermittent job failures due to failing to build bpftrace 


sdimitro:
Tail of log:
```
15:17:17 Err:7 http://apt.llvm.org/bionic llvm-toolchain-bionic/main Sources
15:17:17   File has unexpected size (2527 != 2528). Mirror sync in progress? [IP: 151.101.190.49 80]
15:17:17   Hashes of expected file:
15:17:17    - Filesize:2528 [weak]
15:17:17    - SHA256:1debfc554e873cb9a7551fd1339b571f49d33ba25a9712c84cf3f91d55dc974a
15:17:17    - SHA1:8dc32ec3ac58e6b816f85292bb48deda69162a37 [weak]
15:17:17    - MD5Sum:baed3593f037e02364b7cb4e9bb393a2 [weak]
15:17:17   Release file created at: Mon, 24 Dec 2018 22:26:14 +0000
15:17:17 Hit:8 http://ppa.launchpad.net/ansible/ansible/ubuntu bionic InRelease
15:17:17 Get:9 http://apt.llvm.org/bionic llvm-toolchain-bionic/main amd64 Packages [8166 B]
15:17:17 Get:10 http://apt.llvm.org/bionic llvm-toolchain-bionic-5.0/main Sources [2063 B]
15:17:17 Get:11 http://apt.llvm.org/bionic llvm-toolchain-bionic-5.0/main amd64 Packages [7178 B]
15:17:17 Fetched 25.9 kB in 0s (57.8 kB/s)
15:17:18 Reading package lists...
15:17:18 E: Failed to fetch http://apt.llvm.org/bionic/dists/llvm-toolchain-bionic/main/source/Sources.gz  File has unexpected size (2527 != 2528). Mirror sync in progress? [IP: 151.101.190.49 80]
15:17:18    Hashes of expected file:
15:17:18     - Filesize:2528 [weak]
15:17:18     - SHA256:1debfc554e873cb9a7551fd1339b571f49d33ba25a9712c84cf3f91d55dc974a
15:17:18     - SHA1:8dc32ec3ac58e6b816f85292bb48deda69162a37 [weak]
15:17:18     - MD5Sum:baed3593f037e02364b7cb4e9bb393a2 [weak]
15:17:18    Release file created at: Mon, 24 Dec 2018 22:26:14 +0000
15:17:18 E: Some index files failed to download. They have been ignored, or old ones used instead.
15:17:18 Error: failed command 'sudo apt-get update'
[...]

PR #19, opened 2019-02-22
 Retry multiple times when installing package dependencies

pzakha:
Since we've seen a few issues when fetching packages from external repositories (e.g. #7), this should hopefully help us in case there are intermittent issues.

## TESTING
- linux-pkg-build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/pre-push/96/
- make check
- negative test installing broken dependencies: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/pre-push/117/console

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -406,7 +406,7 @@ function install_pkgs() {
 		echo "apt-get install failed, retrying."
 		sleep 10
 	done
-	echo_error "apt-get install failed after $attempt attempts"
+	die "apt-get install failed after $attempt attempts"
 }
 
 function install_source_package_build_deps() {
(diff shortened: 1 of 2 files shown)


lib/common.sh
```diff
@@ -398,7 +398,15 @@ function get_package_config_from_env() {
 }
 
 function install_pkgs() {
-	logmust sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y "$@"
+	for attempt in {1..3}; do
+		echo "Running: sudo env DEBIAN_FRONTEND=noninteractive " \
+			"apt-get install -y $*"
+		sudo env DEBIAN_FRONTEND=noninteractive apt-get install \
```
sebroy: This is no longer going through logmust. Why not?

pzakha: When logmust fails it kills the script.

But I should have also added a "die" statement after "echo_error"

PR #21, opened 2019-03-07
add Java 8

pzakha:
Add Java 8 package to Linux-pkg.
The main goal here is to stop building Java in appliance-build and do it in Linux-pkg instead. This is also a prerequisite for adding the SSO app to Linux-pkg.

## NEXT STEPS
Remove Java build from appliance-build

## TESTING
* linux-pkg-build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/pre-push/144/
* ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/651/
* make check
* Tested manually building with the SSO app

### DIFF
--- a/packages/java8/config.sh
+++ b/packages/java8/config.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL=none
+
+tarfile="jdk-8u171-linux-x64.tar.gz"
+jdk_path="/usr/lib/jvm/oracle-java8-jdk-amd64"
+
+function prepare() {
+	logmust install_pkgs java-package
+}
+
+function fetch() {
+	logmust cd "$WORKDIR"
+	local url="http://artifactory.delphix.com/artifactory"
+	logmust wget -nv "$url/java-binaries/linux/jdk/8/$tarfile" -O "$tarfile"
+}
+
+function build() {
+	logmust cd "$WORKDIR"
+	logmust env DEB_BUILD_OPTIONS=nostrip fakeroot make-jpkg "$tarfile" <<<y
+	logmust mv ./*.deb "$WORKDIR/artifacts/"
+	#
+	# Store the install path of the JDK in a file so that the users of this
+	# Java package know where to look. This is especially useful for
+	# other linux-pkg packages that have a build dependency on this
+	# particular version of Java, as they don't have to hardcode the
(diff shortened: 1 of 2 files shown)


packages/java8/config.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+#
+# Copyright 2018 Delphix
```
sebroy: nit: 2019

packages/java8/config.sh
```diff
@@ -0,0 +1,40 @@
...
+	logmust cd "$WORKDIR/"
+	local url="http://artifactory.delphix.com/artifactory"
+	logmust wget -nv "$url/java-binaries/linux/jdk/8/$tarfile" -O "$tarfile"
+}
+
+function build() {
+	logmust cd "$WORKDIR/"
+	logmust env DEB_BUILD_OPTIONS=nostrip fakeroot make-jpkg "$tarfile" <<<y
+	logmust mv ./*.deb "$WORKDIR/artifacts/"
+	logmust bash -c "echo $jdk_path >'$WORKDIR/artifacts/JDK_PATH'"
```
prakashsurya: Why do we do this?

packages/java8/config.sh
```diff
@@ -0,0 +1,40 @@
...
+	logmust install_pkgs "$WORKDIR/artifacts/"*.deb
```
prakashsurya: Why do we do this?

pzakha:
@prakashsurya I've added comments in the code that provide more details on why we do certain things.

packages/java8/config.sh
```diff
@@ -0,0 +1,53 @@
...
+	# path in their build definition. This would also be useful if external
+	# packages, such as the app-gate, decide to fetch and install Java from
+	# the Linux-pkg bundle.
+	#
+	logmust bash -c "echo $jdk_path >'$WORKDIR/artifacts/JDK_PATH'"
+	#
+	# Install the Java package on this system so that other linux-pkg
+	# packages can use it.
+	#
+	logmust install_pkgs "$WORKDIR/artifacts/"*.deb
```
prakashsurya: This is an odd way to do build dependencies, but ok..

packages/java8/config.sh
```diff
@@ -0,0 +1,53 @@
...
+	#
+	# Store the install path of the JDK in a file so that the users of this
+	# Java package know where to look. This is especially useful for
+	# other linux-pkg packages that have a build dependency on this
+	# particular version of Java, as they don't have to hardcode the
+	# path in their build definition. This would also be useful if external
+	# packages, such as the app-gate, decide to fetch and install Java from
+	# the Linux-pkg bundle.
+	#
+	logmust bash -c "echo $jdk_path >'$WORKDIR/artifacts/JDK_PATH'"
```
prakashsurya: I still don't fully understand why we do this, and/or how this will be consumed..

Why wouln't consumers just use `/usr/bin/java` which should use the "alternatives" system, so it's a symlink to `/usr/lib/jvm/oracle-java8-jdk-amd64` ?

Will we have multiple versions of java installed, so we can't simply use the "default" version?

packages/java8/config.sh
pzakha: Yeah, we'll probably need to revisit how we do this. For now we are doing this for bcc & bpftrace. I have created https://github.com/delphix/linux-pkg/issues/22 to track this.

packages/java8/config.sh
pzakha: > Will we have multiple versions of java installed, so we can't simply use the "default" version?
That's right, bootstrap-18-04 already has another version of java 8 installed.

PR #38, opened 2019-06-11
Add config.sh for crypt-blowfish packages

prakashsurya:


prakashsurya:
`pre-push` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/129/console)

rasantel:
LGTM. Thanks for taking care of this.

packages/crypt-blowfish/config.sh
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL="https://gitlab.delphix.com/os-platform/linux-3rd-party-pkgs.git"
```
sebroy: What is this URL? Will this be replaced with a path to the actual `crypt-blowfish` repo on gitlab or github?

prakashsurya: This is where the code currently lives. I didn't put it there, so I didn't feel comfortable changing it while also migrating to linux-pkg.

I'd suggest we move forward with this change, to get the existing code integrated with linux-pkg rather than artifactory.. and then, later, we can discuss how/if this repository should be moved.

We can easily move the repository location, and then have a one line change here to use that new location, when it's been moved.

PR #42, opened 2019-07-02
Update current version to 3.4.1

prashks:
Upstream `td-agent` package version was updated to 3.4.1 and I've updated our down stream `td-agent` plugins to match that via - https://github.com/delphix/omnibus-td-agent/pull/4

So, this PR just updates our corresponding `td-agent` package version to 3.4.1 to reflect the same.

### DIFF
--- a/packages/td-agent/config.sh
+++ b/packages/td-agent/config.sh
@@ -17,9 +17,7 @@
 # shellcheck disable=SC2034
 
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/omnibus-td-agent.git"
-
-# We're customizing td-agent version 3.4.1
-DEFAULT_PACKAGE_VERSION=3.4.1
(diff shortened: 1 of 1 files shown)


prashks:
Updated to do this dynamically, see [this diff](https://github.com/delphix/linux-pkg/pull/42/commits/321fdef828e2321788d77c22671213239aaf3935)

Tested  `buildpkg.sh td-agent` and verified it sets `PACKAGE_VERSION` as expected and the final artifact name as expected (`td-agent_3.4.1-delphix-2019.07.02.23_amd64.deb`) -

```PACKAGE td-agent: STAGE build STARTED
Running: build
Running: cd /home/ubuntu/linux-pkg/packages/td-agent/tmp/repo
Running: eval PACKAGE_VERSION="3.4.1"
<<snip>>
```

Issue #27, opened 2019-03-20
Allow re-using previously built packages if no changes were detected


pzakha:
The build should be able to re-use previously built packages if no changes were detected.

This would speed-up ab-pre-push runs and potentially unblock reduce friction when flakey packages do not need to be rebuilt as often.

There are some limitations on when a package can be re-used and it depends on what triggers are used to detect there were no changes. The latest build from `ops.jenkins` would be used as a reference and source for previously built packages.

A package would have to be rebuilt if:
1. linux-pkg's repository hash has changed
2. the package's repository hash has changed
3. The package depends on kernel and the kernel version has changed

Issue #20, opened 2019-02-27
Error: failed command 'bin/omnibus build td-agent3'


prakashsurya:
We hit a failure in the nightly build [here](http://ops.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/post-push/327/console)

The error message shows this:
```
07:35:25 Progress: |    [NetFetcher: ncurses] I | 2019-02-27T15:35:12+00:00 | Retrying failed download due to Net::OpenTimeout (3 retries left)...
07:35:25 
07:36:22 Progress: |    [NetFetcher: ncurses] I | 2019-02-27T15:36:14+00:00 | Retrying failed download due to Net::OpenTimeout (2 retries left)...
07:36:22 
07:37:20 Progress: |    [NetFetcher: ncurses] I | 2019-02-27T15:37:14+00:00 | Retrying failed download due to Net::OpenTimeout (1 retries left)...
07:37:20 
07:38:17 Progress: |    [NetFetcher: ncurses] E | 2019-02-27T15:38:14+00:00 | Download failed - Net::OpenTimeout!
07:38:17 #<Thread:0x00005630f0757a70@/home/ubuntu/.bundle/ruby/2.5.0/omnibus-1cfcf1714d60/lib/omnibus/thread_pool.rb:57 run> terminated with exception (report_on_exception is true):
07:38:17 /usr/lib/ruby/2.5.0/net/http.rb:937:in `initialize': execution expired (Net::OpenTimeout)
07:38:17 	from /usr/lib/ruby/2.5.0/net/http.rb:937:in `open'
07:38:17 	from /usr/lib/ruby/2.5.0/net/http.rb:937:in `block in connect'
07:38:17 	from /usr/lib/ruby/2.5.0/timeout.rb:103:in `timeout'
07:38:17 	from /usr/lib/ruby/2.5.0/net/http.rb:935:in `connect'
07:38:17 	from /usr/lib/ruby/2.5.0/net/http.rb:920:in `do_start'
07:38:17 	from /usr/lib/ruby/2.5.0/net/http.rb:909:in `start'
07:38:17 	from /usr/lib/ruby/2.5.0/open-uri.rb:337:in `open_http'
07:38:17 	from /usr/lib/ruby/2.5.0/open-uri.rb:755:in `buffer_open'
07:38:17 	from /usr/lib/ruby/2.5.0/open-uri.rb:226:in `block in open_loop'
[...]

PR #43, opened 2019-07-02
Allow fetching previously built artifacts from S3

pzakha:
## DESCRIPTION

This gives us the ability to fetch artifacts for some packages from a previous build.

When we do so is dictated by the REUSE_CACHED_ARTIFACTS policy, which can have 3 possible values:
- "never": never re-use artifacts from previous builds. This is the default setting.
- "on-failure": build all packages as usual. If a package fails to build, then fetch it from a previous build.
- "always": fetch all packages from a previous build.

Note that if we pass custom parameters for a package (e.g. provide a custom git url or branch), then we will never re-use artifacts from a previous build.

The "previous build" URL will be passed by Jenkins (devops-gate change to be posted soon) and is the latest post-push build.

Note that linux-pkg will not attempt to compare the hash of previous packages to determine whether or not it should be rebuilt, nor will it care if linux-pkg itself has changed. This simplifies the logic a lot but puts the burden or determining when it is safe to use each setting on the developer that tests the changes.

## MOTIVATION

We've had intermittent external dependency errors such as https://github.com/delphix/linux-pkg/issues/20 and https://github.com/delphix/linux-pkg/issues/7 that prevent developers from testing their changes in other packages. This helps developers to avoid being affected by those errors unless they are making changes to the affected packages.

This will also give the ability to manually launch a post-push job with the "on-failure" setting, to avoid blocking teams that are waiting on some changes to be rolled-out.

By default, git-ab-pre-push will set the "on-failure" setting. As a bonus, we also have the "always" setting, which significantly speeds-up builds. We may decide to use this as the default setting if we deem it safe enough. In any case, we will add an option to pass this setting to git-ab-pre-push.

## TODO
- Add new parameters to documentation

## FOLLOW-UP WORK

- Changes to linux-pkg-build and ab-orchestrator jenkins jobs (tested already, will post soon)
- Changes to git-ab-pre-push (TODO)

## TESTING

- Manually tested various combinations of settings on a bootstrap VM
- Made sure that packages with custom parameters are always built regardless of settings
- Ran linux-pkg-build job with each setting:
  "always": http://collins.d.delphix.com:21968/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/11/consoleFull
  "always", 1 custom package: http://collins.d.delphix.com:21968/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/13
  "on-failure": http://collins.d.delphix.com:21968/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/14/parameters/
  "never": http://collins.d.delphix.com:21968/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/16/parameters/

fixes https://github.com/delphix/linux-pkg/issues/27

### DIFF
--- a/buildlist.sh
+++ b/buildlist.sh
@@ -46,6 +46,7 @@ pkg_list_file="$_RET"
 logmust cd "$TOP"
 logmust make clean
 logmust mkdir artifacts
+logmust mkdir artifacts/cache
 
 #
 # Auto-generate the default revision for all the packages. It will be the
@@ -87,6 +88,19 @@ logmust cp build-info-pkg/artifacts/* artifacts/
 
 for pkg in "${PACKAGES[@]}"; do
 	logmust cp "packages/$pkg/tmp/artifacts"/* artifacts/
+
+	#
+	# Cache each package's artifacts in a separate directory so that they
+	# can be easily retrieved by future linux-pkg builds. Note that those
+	# artifacts are not consumed by appliance-build.
+	#
+	logmust mkdir -p "artifacts/cache/$pkg/artifacts"
+	logmust cp "packages/$pkg/tmp/artifacts"/* \
+		"artifacts/cache/$pkg/artifacts/"
+	if [[ -f "packages/$pkg/tmp/build_info" ]]; then
+		logmust cp "packages/$pkg/tmp/build_info" \
+			"artifacts/cache/$pkg/"
+	fi
 done
 
 echo_success "Packages have been built successfully."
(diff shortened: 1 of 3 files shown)


pzakha:
cc: @prakashsurya @sebroy @jgallag88

pzakha:
Closing this as the main issue was mitigated by https://github.com/delphix/linux-pkg/pull/57.

If we decide to revisit this, we should open a new PR.

PR #44, opened 2019-07-25
add build definition for base-os-tools

pzakha:
Adds the base-os-tools package.
Currently this package doesn't contain anything.

In the future it will be used as a dependency of other packages in the Delphix Appliance and provide functionality such as what is needed for https://jira.delphix.com/browse/DLPX-65226.

## Testing
 - make check
 - linux-pkg-build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/262
 - ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1793

### DIFF
--- a/package-lists/build/userland.pkgs
+++ b/package-lists/build/userland.pkgs
@@ -6,22 +6,23 @@
 # depend on them being built.
 #  - bcc is required by bpftrace
 #  - java8 is required by the saml app
+adoptopenjdk
 bcc
 java8
 make-jpkg
-adoptopenjdk
 
-delphix-sso-app
+base-os-tools
 bpftrace
 challenge-response
 cloud-init
 crash-python
 crypt-blowfish
 delphix-platform
+delphix-sso-app
 drgn
 gdb-python
 libkdumpfile
+ptools
 python-rtslib-fb
 targetcli-fb
-ptools
 td-agent
--- a/packages/base-os-tools/config.sh
+++ b/packages/base-os-tools/config.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/base-os-tools.git"
+DEFAULT_PACKAGE_VERSION=1.0.0
+
+function build() {
+	logmust dpkg_buildpackage_default
+}


pzakha:
cc: @sebroy @grwilson

prakashsurya:
Why wouldn't we use the "delphix-platform" package to contain any required "base OS tools"? We already use that for various package dependencies, and various in-house "tools" (e.g. the script to download the upgrade image), so why wouldn't we want to use that package for any additional tools and/or dependencies?

pzakha:
It just seems like we got some of the dependencies wrong.

Let's say the `challenge_response` package requires some tool (say `/bin/get-system-uuid`). If that tool is provided by `delphix-platform` then we'd have `challenge_response` depend on `dlephix-platform`, however it's currently the opposite: we have `delphix-platform` depend on `challenge_response`, and I don't think we can (or should?) have 2 packages depend on each other.

prakashsurya:
If the goal/requirement is to avoid circular dependencies, then it would have been nice to have that stated in the PR description. Currently, the description doesn't have any information about why we'd want to add this empty package, nor does it give any indication of the long term goal(s) of this new empty package.

IMO, I wouldn't be opposed to circular dependencies like you're suggesting, assuming it'll work correctly for initial install and upgrade. Conceptually, stating that if (any) one package installed, the other needs to be installed as well, seems reasonable to me. I'm not sure if there's any hidden implementations issues that'd arise with this, though; but off the top of my head, I would expect it to work.

If for whatever reason we didn't want to have package dependencies like this, I would probably suggest a more targeted package, specifically for this new tool; e.g. `cra-tools`, `uuid-tools`, or something like that. Otherwise, it's hard for me to determine when to put something in "delphix-platform" vs. "base-os-tools" since both (by design, right?) have a fairly vague definition.

I'm not necessarily opposed to having some package like "base-os-tools", but I'd like us to then make it clear (e.g. with documentation that we can reference) to future developers where the line is between that and "delphix-platform". Based on that line, we might decide to move things from "delphix-platform" to "base-os-tools".

Ultimately, I'm thinking high-level about the design of all our packages, and how they all align with each other, rather than specific details about the package system we're using and its limitations.

With all that said, can you comment on what you're view is on how/where this package fits into the current architecture and why it's useful (ignoring the circular dependency question/problem)?

pzakha:
I'll add more descriptions in the README for why we need this package.

I agree that we'd want to avoid confusion as to when to put things in there vs delphix-platform, so I'll make sure that it is made clear.

Regarding circular dependencies, we should definitely avoid them (see https://www.debian.org/doc/debian-policy/ch-relationships.html, section 7.2, grep for "circular").

prakashsurya:
> I would probably suggest a more targeted package, specifically for this new tool; e.g. cra-tools, uuid-tools

Any comments on this? Why is having a generic "os tools" package better than a more specific package with just this "get system uuid" tool? Alternatively, why couldn't/shouldn't the CRA package itself provide this script, avoiding this package and the dependency problem altogether?

I still have concerns about whether this is the best solution for the stated problem; I feel like it's adding unnecessary complexity.

rasantel:
I'm thinking that perhaps the CRA PAM module should depend on `delphix-platform` and not the other way around. That way `delphix-platform` could provide all essential tools that CRA PAM requires. Specifically, due to DLPX-65106, CRA PAM will now need to get the appliance version. The script `get_appliance_version.sh` exists for that purpose. It's currently in the app gate, and could be moved to `delphix-platform`.

CRA also will need a couple of other shell functions in `dlpx-app-gate/appliance/server/sh/lib/common.sh`. Those functions (or perhaps the entire `common.sh` lib) could be moved to `delphix-platform` as well.

rasantel:
Or we could just keep the dependency the way it is (platform depends on CRA) but at runtime CRA will have to assume the `delphix-platform` package is installed.

pzakha:
Yeah, it seems like we have a bunch of other dependencies that are currently messed up, so adding this script to delphix-platfom won't make it worse.

I'll close this for now, but we should probably revisit what we put in what packages otherwise it will become very confusing.

PR #45, opened 2019-07-25
Store Build Info by default

pzakha:
With this code in place, we can simplify the `config.sh` file for most packages by removing the call to `store_git_info()`. This will also make sure that whenever we build a package from git, the build info is always stored as we forgot to add it for a few packages.

## Testing:
- make check
- verify contents of new global build-info file
- linux-pkg-build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/261/
- ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/1792

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,6 +391,20 @@ should only be overridden when not fetching the package source from git.
 The `build()` hook is mandatory. It is responsible for building the package and
 storing the build products into `packages/<package>/tmp/artifacts/`.
 
+#### Store Build Info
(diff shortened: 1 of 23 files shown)


pzakha:
cc: @sebroy @prakashsurya

PR #46, opened 2019-08-02
Use "libkdumpfile" when building "drgn"

prakashsurya:
We've integrated changes to "drgn" such that it will automatically
detect "libkdumpfile" if it's installed, and enable support for kernel
core dump files when it's detected. To ensure "drgn" is built with this
functionality enabled, we need to install the "libkdumpfile" package
after building it, and prior to building the "drgn" package.

### DIFF
--- a/packages/libkdumpfile/config.sh
+++ b/packages/libkdumpfile/config.sh
@@ -36,4 +36,7 @@ function prepare() {
 
 function build() {
 	logmust dpkg_buildpackage_default
+
+	# Install libkdumpfile, it's needed to build drgn
+	logmust install_pkgs "$WORKDIR/artifacts"/*.deb
 }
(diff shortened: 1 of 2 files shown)


prakashsurya:
I don't know how to test this, it looks like `git-ab-pre-push` is broken:
```
± git-ab-pre-push
Unsupported repository detected: https://github.com/delphix/linux-pkg.git
```

packages/libkdumpfile/config.sh
```diff
@@ -36,4 +36,7 @@ function prepare() {
 
 function build() {
 	logmust dpkg_buildpackage_default
+
+	# Install libkdumpfile, it's needed to build drgn
+	logmust install_pkgs "$WORKDIR/artifacts/*.deb"
```
pzakha: `"$WORKDIR/artifacts/*.deb"` -> `"$WORKDIR"/artifacts/*.deb`
If you quote `*`, it won't be resolved by bash.

prakashsurya: Sorry, right.. Thanks. I need more :coffee:

PR #47, opened 2019-08-05
Use "github.com" hosted "drgn" repository

prakashsurya:


prakashsurya:
- `linux-pkg-build` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/272/)

prakashsurya:
cc: @sdimitro @pzakha

packages/drgn/config.sh
```diff
@@ -16,7 +16,7 @@
 #
 
 # shellcheck disable=SC2034
-DEFAULT_PACKAGE_GIT_URL="https://gitlab.delphix.com/os-platform/drgn.git"
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/drgn.git"
 DEFAULT_PACKAGE_GIT_BRANCH="master"
```
pzakha: This is redundant, "master" is already the default.

prakashsurya: This is unrelated to my changes, so I'm dropping this. I recommend opening a bug and have it addressed in another PR.

prakashsurya:
>  Would be worth it to do a git-ab-pre-push as well once the linux-pkg-build is done.

Yes, I can do this.

> Also, do you have plans to setup auto-update for this repository?

No concrete plans, but I believe we do want to have it be automatically updated. I believe our plans are to use the same model for this repository, as we do for "delphix/zfs"; i.e. we interact directly with upstream, and use auto-merges to keep our for up-to-date.

I don't know how this will be implemented, but I'd like to have that discussion elsewhere, since it's not directly related to this change.

prakashsurya:
@pzakha can you give this another look; I think it's ready to go.

PR #49, opened 2019-08-09
Add config.sh for sdb package

prakashsurya:


prakashsurya:
- `linux-pkg-build` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/274/)

prakashsurya:
cc: @sdimitro @pzakha

I think this is good to go.

packages/sdb/config.sh
```diff
@@ -0,0 +1,30 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# shellcheck disable=SC2034
+DEFAULT_PACKAGE_GIT_URL="https://github.com/sdimitro/sdb.git"
```
pzakha: Can we move this under `delphix` first?

prakashsurya: We can. We loosely decided not to yet, since it's not clear where the final resting place for SDB is going to be. We'd like for it to be in some yet-to-be-created github group, and the delphix fork would be off of that, but since we don't have that group yet, we haven't bothered to create the delphix fork.

If getting the delphix fork is a blocker for this, then we can look into expediting the creation of this github group.

pzakha: I see, so the plan is to make this a fully OSS project?

I won't block this if that makes it significantly harder to iterate over `sdb` without this in linux-pkg, but I definitely don't think it's great to add private repo dependencies to the build. 

We should expedite the creation of the upstream in any case, given that the code complete date is approaching fast.

prakashsurya:
> so the plan is to make this a fully OSS project?

I don't know what you mean by "fully OSS" here; it will be open source regardless of if "upstream" is under our delphix github organization or not.

>  but I definitely don't think it's great to add private repo dependencies to the build.

Generally, I agree. I'm OK blocking this until we get everything sorted out, it just means it's more difficult for folks to use it, while we're still working on it (folks have to build/install from source, rather than having it preinstalled).

pzakha:
> I'm OK blocking this until we get everything sorted out, it just means it's more difficult for folks to use it, while we're still working on it

Alright, I'll leave this decision up to you.

As an alternative to make it easier you can also push changes to a private branch of yours, create a build list with just `sdb`, and then use http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/custom/job/pre-push/ to build it. This way anyone working on this will be able to just rerun that job to build the package.

prakashsurya:
The SDB repository is now hosted under our Delphix organization; I've updated this PR and restarted a new build of `linux-pkg-build`.

prakashsurya:
@pzakha the build succeeded, so I think it's ready to be integrated.

PR #51, opened 2019-09-03
DLPX-64156 crashdump should only include ZFS metadata

don-brady:
**Description**

Add `makedumpfile` package

We need a modified `makedumpfile(8)` command that can filter ZFS file data pages from the kernel crash dump file. This package is one part of the necessary changes to bring about ZFS file data page filtering.

**Related Changes**
The makedumpfile patch is at https://github.com/delphix/makedumpfile/commit/9d7983e44b which adds the `'--private-page-filter'` option

A corresponding ZFS kernel module changes landed in: 
 https://github.com/zfsonlinux/zfs/commit/28c91ab66d

A follow up change, once this patch has landed, is to change the /etc/default/kdump-tools configuration file to start passing the `--private-page-filter` option to makedumpfile.

**Testing**
Appliance build pre-push (using linux-pkg userland build artifacts):
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/2061/

Manually tested using `git-linux-pkg build` to install makedumpfile on a target host and confirmed that it can filter ZFS pages  with the '--private-page-filter' option.

### DIFF
--- a/package-lists/build/userland.pkgs
+++ b/package-lists/build/userland.pkgs
@@ -21,6 +21,7 @@ crypt-blowfish
 delphix-platform
 drgn
 gdb-python
+makedumpfile
 nfs-utils
 python-rtslib-fb
 sdb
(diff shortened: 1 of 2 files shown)


don-brady:
fixed merge conflict

PR #58, opened 2019-09-20
LB-215 Want linux-pkg-build to use package mirror

grodr:
Update linux-pkg-build setup to add the internal linux package mirror as an apt source before doing the build.

Testing:
Ran linux-pkg-build kernel 
http://collins.d.delphix.com:35379/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/kernel/job/post-push/26/

Ran linux-pkg-build userland
http://collins.d.delphix.com:35379/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/post-push/1/

With the output from the above, I ran ab-pre-push 
http://collins.d.delphix.com:35379/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/47/parameters/
and cloned the resulting VM and compared installed packages to a current trunk VM via `dpkg-query -l`, I noted that the only differences were build versions that contain dates of some packages, as my build was from two days ago.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -22,6 +22,30 @@ export DEBIAN_FRONTEND=noninteractive
 # TODO: allow updating upstream for other branches than master
 export REPO_UPSTREAM_BRANCH="upstreams/master"
 
+#
+# Determine DEFAULT_GIT_BRANCH. If it is unset, default to the branch set in
+# branch.config.
+#
+if [[ -z "$DEFAULT_GIT_BRANCH" ]]; then
+	echo "DEFAULT_GIT_BRANCH is not set."
+	if ! source "$TOP/branch.config" 2>/dev/null; then
+		echo "No branch.config file found in repo root."
+		exit 1
+	fi
+
+	if [[ -z "$DEFAULT_GIT_BRANCH" ]]; then
+		echo "$DEFAULT_GIT_BRANCH parameter was not sourced from " \
+			"branch.config. Ensure branch.config is properly formatted with " \
+			"e.g. DEFAULT_GIT_BRANCH=\"<upstream-product-branch>\""
+		exit 1
+	fi
+
+	echo "Defaulting DEFAULT_GIT_BRANCH to branch $DEFAULT_GIT_BRANCH set in" \
+		"branch.config."
+
+	export DEFAULT_GIT_BRANCH
+fi
+
 # shellcheck disable=SC2086
 function enable_colors() {
 	[[ -t 1 ]] && flags="" || flags="-T xterm"
--- a/setup.sh
+++ b/setup.sh
@@ -18,7 +18,48 @@
 TOP="$(dirname "$(readlink -f "${BASH_SOURCE[0]}")")"
 source "$TOP/lib/common.sh"
 
+#
+# Update the sources.list file to point to our internal package mirror. If no
+# mirror url is passed in, then the latest mirror snapshot is used.
+#
+configure_apt_sources() {
+	local package_mirror_url=''
+	if [[ -n "$DELPHIX_PACKAGE_MIRROR_MAIN" ]]; then
+		package_mirror_url="$DELPHIX_PACKAGE_MIRROR_MAIN"
+	else
+		local latest_url="http://linux-package-mirror.delphix.com/"
+		latest_url+="${DEFAULT_GIT_BRANCH}/latest/"
+		package_mirror_url=$(curl -LfSs -o /dev/null -w '%{url_effective}' \
+			"$latest_url" || die "Could not curl $latest_url")
+
+		package_mirror_url+="ubuntu"
+	fi
+
+	#
+	# Remove other sources in sources.list.d if they are present.
+	#
+	[[ -d /etc/apt/sources.list.d ]] && (
+		logmust sudo rm -rf /etc/apt/sources.list.d ||
+			die "Could not remove /etc/apt/sources.list.d"
+	)
+
+	sudo bash -c "cat <<-EOF >/etc/apt/sources.list
+deb ${package_mirror_url} bionic main restricted universe multiverse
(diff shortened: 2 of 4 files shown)


lib/common.sh
```diff
@@ -741,3 +741,44 @@ function determine_target_kernels() {
 	echo "Kernel versions to use to build modules:"
 	echo "  $KERNEL_VERSIONS"
 }
+
+#
+# Update the sources.list file to point to our internal package mirror. If no
+# mirror url is passed in, then the latest mirror snapshot is used.
+#
+configure_apt_sources() {
```
sebroy: This logic is going to need to be shared between `linux-pkg` and `appliance-build`, right? Could we instead require that the package mirror url always be passed in, and put this logic in common devops code to be passed in via jenkins job?

grodr: I have the PR for appliance-build here https://github.com/delphix/appliance-build/pull/365 I wanted this logic at the appliance-build and linux-pkg level rather than the jenkins level since these workflows can be can be ran locally without jenkins, so I wanted to support that workflow. However, this logic could be done at the jenkins. Thoughts?

lib/common.sh
```diff
@@ -741,3 +741,44 @@ function determine_target_kernels() {
...
+    local package_mirror_url=''
+    if [[ -n "$DELPHIX_PACKAGE_MIRROR_MAIN" ]]; then
+        package_mirror_url="$DELPHIX_PACKAGE_MIRROR_MAIN"
+    else
+        local latest_url="http://linux-package-mirror.delphix.com/"
+        latest_url+="${LINUX_PKG_GIT_BRANCH}/latest/"
+        package_mirror_url=$(curl -Ls -o /dev/null -w '%{url_effective}' \
```
jgallag88: The `-S` option would be useful to add. It causes errors which would otherwise be suppressed by `-s` to be printed to stderr.

lib/common.sh
```diff
@@ -741,3 +741,44 @@ function determine_target_kernels() {
...
+# Update the sources.list file to point to our internal package mirror. If no
+# mirror url is passed in, then the latest mirror snapshot is used.
+#
+configure_apt_sources() {
+    local package_mirror_url=''
+    if [[ -n "$DELPHIX_PACKAGE_MIRROR_MAIN" ]]; then
+        package_mirror_url="$DELPHIX_PACKAGE_MIRROR_MAIN"
+    else
+        local latest_url="http://linux-package-mirror.delphix.com/"
+        latest_url+="${LINUX_PKG_GIT_BRANCH}/latest/"
```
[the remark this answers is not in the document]
grodr: Updated to use DEFAULT_GIT_BRANCH.

lib/common.sh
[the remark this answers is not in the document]
grodr: Updated to do `|| die..`
I also updated the curl command to use the -f and -S options so that it will fail if no URL is found.

lib/common.sh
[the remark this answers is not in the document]
grodr: Makes sense, moved.

lib/common.sh
```diff
@@ -741,3 +741,44 @@ function determine_target_kernels() {
 	echo "Kernel versions to use to build modules:"
 	echo "  $KERNEL_VERSIONS"
 }
+
```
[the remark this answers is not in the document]
grodr: I didn't realize this was available. Super convenient! Done.

lib/common.sh
```diff
@@ -741,3 +741,44 @@ function determine_target_kernels() {
...
+            die "No package mirror dataset found at $latest_url"
+        fi
+
+        package_mirror_url+="ubuntu"
+    fi
+
+    #
+    # Remove other sources in sources.list.d if they are present.
+    #
+    [[ -d /etc/apt/sources.list.d ]] && sudo rm -rf /etc/apt/sources.list.d
```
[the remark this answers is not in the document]
grodr: Updated to use logmust

lib/common.sh
```diff
@@ -741,3 +741,44 @@ function determine_target_kernels() {
...
+
+    sudo bash -c "cat <<-EOF >/etc/apt/sources.list
```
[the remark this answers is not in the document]
grodr: Added an `|| die..`

lib/common.sh
grodr: 👍 added the `-S` and `-f` flags to make errors more apparent.

setup.sh
```diff
@@ -18,7 +18,73 @@
...
+			"branch.config. Ensure branch.config is properly formatted with " \
+			"e.g. DEFAULT_GIT_BRANCH=\"<upstream-product-branch>\""
+		exit 1
+	fi
+
+	echo "Defaulting DEFAULT_GIT_BRANCH to branch $DEFAULT_GIT_BRANCH set in" \
+		"branch.config."
+fi
+
+export DEFAULT_GIT_BRANCH="$DEFAULT_GIT_BRANCH"
```
pzakha: This logic should go into `common.sh`

lib/common.sh
```diff
@@ -22,6 +22,30 @@ export DEBIAN_FRONTEND=noninteractive
...
+		echo "$DEFAULT_GIT_BRANCH parameter was not sourced from " \
+			"branch.config. Ensure branch.config is properly formatted with " \
+			"e.g. DEFAULT_GIT_BRANCH=\"<upstream-product-branch>\""
+		exit 1
+	fi
+
+	echo "Defaulting DEFAULT_GIT_BRANCH to branch $DEFAULT_GIT_BRANCH set in" \
+		"branch.config."
+
+	export DEFAULT_GIT_BRANCH="$DEFAULT_GIT_BRANCH"
```
pzakha: nit: `=...` is not needed here and simply `export DEFAULT_GIT_BRANCH` is sufficient since that variable is already set.

grodr: updated

PR #60, opened 2019-10-03
Output debug symbol (.ddeb) packages

jgallag88:
linux-pkg already creates `.ddeb` packages where appropriate, we just need to copy them to the `artifacts/` directory so that they get uploaded with the rest of the packages by Jenkins at the end of a build.

This mostly just involves changing `mv ./*.debs ...` to `mv ./*debs ...` so that the glob will match `.ddeb` packages as well. Note that I've made this change in a few places where it isn't strictly necessary (in packages that don't produce a corresponding `ddeb` package), because it doesn't hurt anything and it helps to establish a pattern to be followed in the future.

### DIFF
--- a/packages/adoptopenjdk/config.sh
+++ b/packages/adoptopenjdk/config.sh
@@ -21,7 +21,7 @@ tarfile="OpenJDK8U-jdk_x64_linux_hotspot_8u212b04.tar.gz"
 jdk_path="/usr/lib/jvm/adoptopenjdk-java8-jdk-amd64"
 
 function prepare() {
-	if ! ls "$TOP/packages/make-jpkg/tmp/artifacts/"*.deb >/dev/null 2>&1; then
+	if ! ls "$TOP/packages/make-jpkg/tmp/artifacts/"*deb >/dev/null 2>&1; then
 		echo_bold "custom java-package not installed. Building package 'make-jpkg' first."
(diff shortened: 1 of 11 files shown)


jgallag88:
### Testing

Checked that all the `ddeb` packages that we want were uploaded at the end of a build.

userland package list build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/351/console
kernel package list build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/kernel/job/pre-push/87/console

packages/adoptopenjdk/config.sh
```diff
@@ -55,7 +55,7 @@ function build() {
 	# Install the Java package on this system so that other linux-pkg
 	# packages can use it.
 	#
-	logmust install_pkgs "$WORKDIR/artifacts/"*.deb
+	logmust install_pkgs "$WORKDIR/artifacts/"*deb
```
pzakha: Here and elsewhere that does `install_pkgs`, no need to install `.ddeb` on the bootstrap VM.

jgallag88: Fixed.

PR #61, opened 2019-10-07
drgn: install build dependencies using info from control file

sdimitro:
depends on: https://github.com/delphix/drgn/pull/3

Testing:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/356/

userland-job without the pkg-changes:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/357/

### DIFF
--- a/packages/drgn/config.sh
+++ b/packages/drgn/config.sh
@@ -23,24 +23,17 @@ UPSTREAM_GIT_URL="https://github.com/osandov/drgn.git"
 UPSTREAM_GIT_BRANCH="master"
 
 function prepare() {
-	logmust install_pkgs \
-		autoconf \
-		automake \
-		bison \
-		flex \
-		gawk \
-		git \
-		gcc \
-		libbz2-dev \
-		libdw-dev \
-		libelf-dev \
-		liblzma-dev \
-		libtool \
-		make \
-		pkg-config \
-		python3-distutils \
-		python3.6-dev \
-		zlib1g-dev
+	#
+	# Strictly speaking libkdumpfile is not a hard prerequisite for
+	# drgn itself, but it is a hard requirement in our use-case as
+	# we do want to use drgn for kdump-compressed crash dumps.
+	#
+	if ! dpkg-query --show libkdumpfile >/dev/null 2>&1; then
+		echo_bold "libkdumpfile not installed. Building package 'libkdumpfile' first."
+		logmust "$TOP/buildpkg.sh" libkdumpfile
+	fi
+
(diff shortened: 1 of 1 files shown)


sdimitro:
cc: @pzakha @prakashsurya

packages/drgn/config.sh
```diff
@@ -23,24 +23,17 @@ UPSTREAM_GIT_URL="https://github.com/osandov/drgn.git"
...
-		zlib1g-dev
+	#
+	# Strictly speaking libkdumpfile is not a hard prerequisite for
+	# drgn itself, but it is a hard requirement in our use-case as
+	# we do want to use drgn for kdump-compressed crash dumps.
+	#
+        if ! dpkg-query --show libkdumpfile >/dev/null 2>&1; then
+                echo_bold "libkdumpfile not installed. Building package 'libkdumpfile' first."
+                logmust "$TOP/buildpkg.sh" libkdumpfile
+        fi
```
pzakha: Do we need `libkdumpfile` to be installed on the build machine for properly building `drgn`? If not, then we probably do not need this code.

sdimitro: `drgn`'s install scripts check for `libkdumpfile` when compiling the code and if the library exists the some `#define`s are enabled. These `#define`s enable `drgn` to be compiled with support for compressed crash-dumps which we use within Delphix.

Unfortunately this functionality is not something that we can enable at `drgn`'s runtime (at least not yet), thus I wrote this the following way to enforce it as a build dependency.

Does this make sense?

pzakha: I see. Yeah, that makes sense, thanks for the explanation. In this case this change looks good.

packages/drgn/config.sh
```diff
@@ -23,24 +23,17 @@ UPSTREAM_GIT_URL="https://github.com/osandov/drgn.git"
...
+
+        logmust install_build_deps_from_control_file
```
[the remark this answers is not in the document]
sdimitro: fixed

sdimitro:
userland-build job with both changes succeeded, see link in description.

sdimitro:
running the userland-job without the linux-pkg changes seems to be successful too: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/357/

PR #66, opened 2019-11-14
Add performance-diagnostics package

jgallag88:
Create performance-diagnostics package and add it to `userland` list. Note that this shouldn't be merged until `DEFAULT_PACKAGE_GIT_URL` has been updated, which can be done after the performance-diagnostics code is pushed to the official repo.

### DIFF
--- a/packages/performance-diagnostics/config.sh
+++ b/packages/performance-diagnostics/config.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+#
+# Copyright 2019 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+#
+
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/performance-diagnostics.git"
+DEFAULT_PACKAGE_VERSION="1.0.0"
+
+function prepare() {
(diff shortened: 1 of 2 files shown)


jgallag88:
linux-pkg userland build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/377/console

packages/performance-diagnostics/config.sh
```diff
@@ -0,0 +1,31 @@
...
+#
+# shellcheck disable=SC2034
+#
+
+# TODO point to official repo
+DEFAULT_PACKAGE_GIT_URL="https://gitlab.delphix.com/john.gallagher/perf-playbook-bcc.git"
+DEFAULT_PACKAGE_VERSION="1.0.0"
+
+function prepare() {
+	logmust install_pkgs devscripts
```
pzakha: We should use `logmust install_build_deps_from_control_file` whenever possible.
This way we do not need to update this file whenever we add new build dependencies.

packages/performance-diagnostics/config.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+function build() {
+	logmust cd "$WORKDIR/repo"
```
pzakha: nit: this is not needed.

packages/performance-diagnostics/config.sh
jgallag88: Looking at this more, I think this can actually be removed entirely. `devscripts` is really more of a dependency of `linux-pkg`, and `linux-pkg` is already installing it in `setup.sh`.

pzakha: Sure, I'd still add `logmust install_build_deps_from_control_file` here since if we add more build dependencies in the future then we'll be covered.

pzakha: That said if we don't think this package will need any build dependencies I'm fine with this.

jgallag88: Makes sense to be safe

jgallag88:
new linux-pkg build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/390/console

PR #73, opened 2019-12-16
Add "-r" option to "updatelist.sh" script

prakashsurya:
This change adds the "-r" option to the "updatelist.sh" script to enable
us to update only the packages that pull from source packages. This is
useful for updating release branches with important security updates
that may come from upstream source packages.

### DIFF
--- a/updatelist.sh
+++ b/updatelist.sh
@@ -58,6 +58,7 @@ function usage() {
 	echo "package-lists/update/<list>.pkgs."
 	echo ""
 	echo "    -n  dry-run. Pass the dry-run flag to git push (git push -n)."
+	echo "    -r  release. Update packages for a release branch."
 	echo "    -s  stop on failure. By default script continues when update"
 	echo "        for a package fails."
 	echo "    -h  display this message and exit."
@@ -65,10 +66,12 @@ function usage() {
 	exit 2
 }
 
+release=false
 dry_run=false
 stop_on_failure=false
-while getopts ':hns' c; do
+while getopts ':hnrs' c; do
 	case "$c" in
+	r) release=true ;;
 	s) stop_on_failure=true ;;
 	n) dry_run=true ;;
 	h) usage >&2 ;;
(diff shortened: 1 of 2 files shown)


prakashsurya:
cc: @sebroy @shartse

This would allow us to limit updates for releases to only those packages that pull from an upstream source package (which will come from our Ubuntu APT mirror).

We'd have to add logic to the devops-gate to do the updates for our release branches, and pass in this new flag.

I'd also need to make sure the update logic is smart enough to push to (and update) the release branches of each project, rather than "master".. since we don't currently update release branch, I can't assume that works.

prakashsurya:
> I'd also need to make sure the update logic is smart enough to push to (and update) the release branches of each project, rather than "master".. since we don't currently update release branch, I can't assume that works.

```
 22 # TODO: allow updating upstream for other branches than master                  
 23 export REPO_UPSTREAM_BRANCH="upstreams/master" 
```
and now, looking a little closer.. it looks like it does not yet support updating non-master branches.. so we'd also need to add that support.

updatelist.sh
```diff
@@ -58,17 +58,20 @@ function usage() {
 	echo "package-lists/update/<list>.pkgs."
 	echo ""
 	echo "    -n  dry-run. Pass the dry-run flag to git push (git push -n)."
+	echo "    -r  release. Update packages for a release branch."
```
delphix-release-admin: I think this naming is a little ambiguous - we'll probably mostly want to update packages on the `/stage` branch. This is technically a "release" branch but not the `/release` branch. I'll try and think of something more precise.

shartse: I think this naming is a little ambiguous - we'll probably mostly want to update packages on the /stage branch. This is technically a "release" branch but not the /release branch. I'll try and think of something more precise.

prakashsurya: Yea, I agree. I didn't worry too much about the name yet, we can change it if you have some suggestions.

PR #74, opened 2019-12-27
temporarily disable auto-update of crash package

pzakha:
See discussion on slack.


shartse:
I added `delphix-devops-bot` as a collaborator to `delphix/crash` and then accepted its invite. Let me know if it needs more permissions.

PR #76, opened 2020-01-06
DLPX-67915 Add oracle-specific kernel

grwilson:
This change introduces an oracle-specific kernel and remove the kvm-specific kernel which is no longer used. The kvm-specific kernel is no longer used by any of our platforms and the kvm platform now relies on the generic kernel (see https://github.com/delphix/delphix-platform/pull/151)

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 #
-# Copyright 2018, 2019 Delphix
+# Copyright 2018, 2020 Delphix
 #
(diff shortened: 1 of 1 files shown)


grwilson:
Manually built linux-pkg and delphix-kernel: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/kernel/job/pre-push/123/

The `delphix-kernel` change can be found here: https://github.com/delphix/delphix-kernel/pull/10

PR #85, opened 2020-04-03
Add grub2 package to the Delphix build process

pcd1193182:
This patch adds the grub2 package to the build process. This package recently had its own official repo added to the Delphix organization, and the patches to enable the recovery environment work there are still under review.

Tested via manual builds and installation of the produced packages, and also via automation:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/568/ and http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/kernel/job/pre-push/150/ .  Note that the latter build failed; this is because our version of grub2 doesn't yet have the patch that fixes the build to work on ubuntu with a custom version of ZFS installed. I will update with a new run when that patch is integrated.

### DIFF
--- a/package-lists/build/kernel.pkgs
+++ b/package-lists/build/kernel.pkgs
@@ -3,6 +3,11 @@
 # kernel versions.
 #
 
+# Note: The following packages should be built first because other packages
+# depend on them being built.
+#  - zfs is required by grub2
+zfs
+
 connstat
 delphix-kernel
-zfs
+grub2
--- a/package-lists/update/userland.pkgs
+++ b/package-lists/update/userland.pkgs
@@ -21,3 +21,4 @@ nfs-utils
 python-rtslib-fb
 targetcli-fb
 td-agent
+grub2
--- a/packages/grub2/config.sh
+++ b/packages/grub2/config.sh
@@ -17,9 +17,7 @@
 # shellcheck disable=SC2034
 
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/grub2"
-DEFAULT_PACKAGE_GIT_BRANCH="applied/ubuntu/bionic-updates"
 
-DEFAULT_PACKAGE_VERSION="2.02-2ubuntu8.15"
 UPSTREAM_SOURCE_PACKAGE=grub2
 
 #
@@ -39,6 +37,11 @@ function prepare() {
 # Build the package.
 #
 function build() {
+	logmust cd "$WORKDIR/repo"
+	if [[ -z "$PACKAGE_VERSION" ]]; then
+		logmust eval PACKAGE_VERSION="$(dpkg-parsechangelog -S Version | \
+		    awk -F'-' '{print $1}')"
+	fi
 	logmust dpkg_buildpackage_default
 }
 
--- a/packages/grub2/config.sh
+++ b/packages/grub2/config.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+#
+# Copyright 2020 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/grub2"
+DEFAULT_PACKAGE_GIT_BRANCH="applied/ubuntu/bionic-updates"
+
+DEFAULT_PACKAGE_VERSION="2.02-2ubuntu8.15"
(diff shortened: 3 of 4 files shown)


pzakha:
The change LGTM. Could you please add a section detailing the testing you did?

packages/grub2/config.sh
```diff
@@ -0,0 +1,51 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/grub2"
+DEFAULT_PACKAGE_GIT_BRANCH="applied/ubuntu/bionic-updates"
```
pzakha, line 20: We shouldn't use `DEFAULT_PACKAGE_GIT_BRANCH`, just leave this empty

packages/grub2/config.sh
```diff
@@ -0,0 +1,51 @@
...
+
+DEFAULT_PACKAGE_VERSION="2.02-2ubuntu8.15"
```
pzakha, line 22: We should try to avoid hardcoding the package version in linux-pkg, and instead figure it out from the code of the package being build. See `python-rtslib-fb/config.sh` or `cloud-init/config.sh` as examples.

packages/grub2/config.sh
```diff
@@ -39,6 +37,11 @@ function prepare() {
 # Build the package.
 #
 function build() {
+	logmust cd "$WORKDIR/repo"
+	if [[ -z "$PACKAGE_VERSION" ]]; then
+		logmust eval PACKAGE_VERSION="$(git describe --abbrev=0 2>/dev/null | \
```
pzakha, line 42: I don't think this will work since we update the package from the Ubuntu source package, not from git. As such, the git metadata will not be queried.

It looks like the package version can be retrieved from the `configure` file:
```
PACKAGE_VERSION='2.02'
```

Alternatives for Ubuntu source packages, we may want to provide a helper function that parses the existing changelog. It would be something like:
```
dpkg-parsechangelog -S Version | awk -F'-' '{print $1}'
```

pcd1193182: Yeah I was thinking about the latter option but felt it was a little too hacky. But if we're only pulling the source, I think that's actually the only option that will work.

packages/grub2/config.sh
```diff
@@ -0,0 +1,53 @@
...
+	return
+}
+
+#
+# Build the package.
+#
+function build() {
+	logmust cd "$WORKDIR/repo"
+	if [[ -z "$PACKAGE_VERSION" ]]; then
+		logmust eval PACKAGE_VERSION="$(dpkg-parsechangelog -S Version)"
```
pzakha, line 42: You may want to include `awk -F'-' '{print $1}'` in here. See similar comment here: https://github.com/delphix/linux-pkg/pull/86#discussion_r405499018

package-lists/build/kernel.pkgs
```diff
@@ -3,6 +3,11 @@
 # kernel versions.
 #
 
+# Note: The following packages should be built first because other packages
+# depend on them being built.
+#  - zfs is required by grub2
+zfs
+
```
sebroy, line 10: nit/question: is the blank line between zfs and connstat deliberate?

package-lists/update/userland.pkgs
```diff
@@ -21,3 +21,4 @@ nfs-utils
 python-rtslib-fb
 targetcli-fb
 td-agent
+grub2
```
sebroy, line 24: What does it mean for a package to be contained in both the `kernel.pkgs` and `userland.pkgs` lists?

package-lists/build/kernel.pkgs
pcd1193182: Yeah, the blank line is there to indicate that the first set of packages is being built deliberately before the later ones for build dependency reasons. We do a similar thing in build/userland.pkgs

package-lists/update/userland.pkgs
pcd1193182: One of the updates is in the build/kernel.pkgs file, while the other is in the update/userland.pkgs file. This means that the grub2 bits will be built by the kernel build job, but updated by the userland build job. I don't know of any major downsides to having it be configured this way, which is why I did it, but perhaps @pzakha can give advice if there would be some other configuration that would be preferable.

pzakha: To elaborate on that, when we build the kernel build list, it builds `grub2` along with the other kernel packages because grub2 requires libzfs and that's the way we can do dependencies between linux-pkg packages right now.

Regarding the update list, I had just named it userland because it had userland packages, but really what we should have named it is something like "main.pkgs", which auto-updates all the third-party packages except zfs, since zfs is VIP and runs on a 15-minute polling basis.

sebroy: Got it, thanks for clarifying.

pcd1193182:
http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3336/console

PR #86, opened 2020-04-08
DLPX-69161 [Backport of DLPX-69137 6.0.2.0] walinuxagent package needs to be updated

grwilson:
Testing:

Userland build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/userland/job/pre-push/569/console

Kernel build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg-build/job/master/job/kernel/job/pre-push/153/console

Appliance build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/3244/console

### DIFF
--- a/packages/walinuxagent/config.sh
+++ b/packages/walinuxagent/config.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+#
+# Copyright 2020 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


packages/walinuxagent/config.sh
```diff
@@ -0,0 +1,42 @@
...
+
+function checkstyle() {
+	logmust cd "$WORKDIR/repo"
+	logmust make style-check
+}
+
+function build() {
+	logmust cd "$WORKDIR/repo"
+	PACKAGE_VERSION=$(
+		dpkg-parsechangelog | sed -rne 's,^Version: (.*),\1,p'
```
pzakha, line 35: You can use `dpkg-parsechangelog -S Version | awk -F'-' '{print $1}'` instead.
Note that `dpkg-parsechangelog -S Version` prints the "full version", which is "Version-Revision". Given that we provide our own revision based on a timestamp, we need the awk step. You can see in your build link that it produced `walinuxagent_2.2.45-0ubuntu1~18.04.1-delphix-2020.04.08.01_amd64.deb`, which suggests the full version string would be `2.2.45-0ubuntu1~18.04.1-delphix-2020.04.08.01`. That actually still work, but is quite a long version string.

PR #88, opened 2020-04-13
LB-256 Use github actions to perform sanity checks on pull-requests for linux-pkg

pzakha:
Add github actions to linux-pkg to perform checkstyle.
In the future we will add more sanity checks to linux-pkg PRs.

## Testing
Made sure that the new actions pass on master

### DIFF
--- a/.github/scripts/install-shellcheck.sh
+++ b/.github/scripts/install-shellcheck.sh
@@ -1,0 +1,4 @@
+#!/bin/bash -ex
+
+apt-get update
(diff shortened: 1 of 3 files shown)


prakashsurya:
FWIW, the title of this PR made me think we were going to run "checkstyle" from the individual package configs.. IIRC, we have a "checkstyle" function that can be set in the package configuration, but this PR is not doing that.

I'm not sure if it matters enough to reword the title to be more clear, but just wanted mention this.

PR #91, opened 2020-04-27
Fix shellfmt error

pcd1193182:


pzakha:
This is a dup of https://github.com/delphix/linux-pkg/pull/90

PR #94, opened 2020-05-13
delphix-platform: "make check" is no longer supported

prakashsurya:


prakashsurya:
This is required prior to landing: https://github.com/delphix/delphix-platform/pull/220

PR #97, opened 2020-06-16
Build and update kernel variants through linux-pkg

sdimitro:
Build and update kernel variants through linux-pkg

    = Motivation

    The Ubuntu kernels have served us well but we currently have two
    problems with them:

    1] When we trigger a kernel bug we have to wait until a patch is
       released by Canonical. This may sound like a rare event but
       we've already hit bugs in iSCSI and NFS subsystems that we've
       had to work around.

    2] There are certain kernel config options that we'd like to enable
       that won't be enabled by Canonical until later releases or at
       all even. Examples include BTF info in vmlinux and the PSI api
       needed by oomd.

    = This Patch

    Enable linux-pkg to compile the Ubuntu kernels from scratch. In
    addition it allows us to base our own patches on top of these
    kernels, and rebase them whenever we get new commits from the
    upstream Canonical kernels. This allows us to:

    1] Fix upstream kernel bugs in a timely manner during customer and
       internal escalations.

    2] Customize our kernel config as we see fit.

    3] Have at least one standard workflow for experimenting with
       our own kernel changes that we may want to potentially upstream.

### DIFF
--- a/default-package-config.sh
+++ b/default-package-config.sh
@@ -34,3 +34,269 @@ function store_build_info() {
 		echo "No build info available" >"$WORKDIR/build_info"
 	fi
 }
+
+#
+# The functions below are specific for the Linux kernel packages
+# and contain the majority of their common code.
+#
+function kernel_prepare() {
+	logmust install_pkgs \
+		equivs \
+		devscripts \
+		kernel-wedge
+}
+
+#
+# The configuration disabled below is specifically for uses
+# of ${debian_rules_args}. Quoting the specific variable
+# results in incorrect behavior and thus we disable that
+# check.
+#
+# shellcheck disable=SC2086
+function kernel_build() {
+	local platform="$1"
+	#
+	# Note: Extra arguments can overwrite default arguments.
+	#       For example in this function we default skipdbg
+	#       to false, but if we pass "skipdbg=true" as an
+	#       extra argument we will be overwriting this value
+	#       to true. This is because when a variable's value
+	#       is declared multiple times when invoking the
+	#       debian/rules command, the rightmost declaration
+	#       is the one that is actually used.
+	#
+	local debian_rules_extra_args="$2"
+
+	logmust cd "$WORKDIR/repo"
+
+	#
+	# We generate the default control file from Canonical
+	# so we can capture the ABI number (abinum) from
+	# Canonical's kernel - (see comment that follows for
+	# the reason and the relevant code for the logic).
+	#
+	logmust debian/rules debian/control
+
+	#
+	# We overwrite the default abinum build variable with our
+	# version strings and at the same time retain the original
+	# abinum from Canonical by appending it at the end.
+	#
+	# We chose to mutate the abinum field as it is the least
+	# invasive for Ubuntu's build logic (e.g. most of the other
+	# variables actually interact with logic in the build). At
+	# the same time the abinum variable is part of the fields
+	# that we care about (e.g. package name, linux image file
+	# name, etc..).
+	#
+	# We still retain the original abinum by appending it at
+	# the end of the new one to maintain the mapping between
+	# Canonical's releases and our releases.
+	#
+	local canonical_abinum delphix_abinum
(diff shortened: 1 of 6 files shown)


sdimitro:
cc @pzakha - not expecting a thorough review. This is mostly to demonstrate progress

default-package-config.sh
```diff
@@ -34,3 +34,107 @@ function store_build_info() {
...
+	#   the line.
+	#
+	local debian_rules_args="skipdbg=false uefi_signed=false abinum=${delphix_abinum} ${debian_rules_extra_args}"
+
+	#
+	# Clean up everything generated so far and recreate the
+	# final control file with the arguments that we want.
+	#
+	logmust fakeroot debian/rules clean
+	logmust debian/rules debian/control ${debian_rules_args}
```
pzakha, line 117: Could you just run ```logmust fakeroot debian/rules clean ${debian_rules_args}```?
`clean` ends up running `debian/control`

sdimitro: 👍

sdimitro:
@pzakha updated the PR with the second milestone

default-package-config.sh
```diff
@@ -34,3 +34,268 @@ function store_build_info() {
...
+
+	#
+	# Ensure that there is a commit marking the start of
+	# the Delphix set of patches. Then get the hash of
+	# the commit right before it.
+	#
+	local dlpx_patch_end dlpx_patch_start current_ubuntu_commit
+	dlpx_patch_start=$(git log --pretty=oneline repo-HEAD | grep @@DELPHIX_PATCHSET_START@@ | awk '{ print $1 }')
+	[ -z "${dlpx_patch_start}" ] && die "could not find DELPHIX_PATCHSET_START"
+	current_ubuntu_commit=$(git log --pretty=oneline repo-HEAD | grep -A 1 @@DELPHIX_PATCHSET_START@@ | tail -n 1 | awk '{ print $1 }')
```
pzakha, line 272: nit: You can also use `current_ubuntu_commit=$(git rev-parse ${dlpx_patch_start}^)`

default-package-config.sh
```diff
@@ -34,3 +34,268 @@ function store_build_info() {
...
+	# df8fd7d8802d59   refs/tags/Ubuntu-oracle-5.3-5.3.0-1015.16_18.04.1
+	# 0fe5cd29e90a5e   refs/tags/Ubuntu-oracle-5.3-5.3.0-1015.16_18.04.2
+	# ```
+	#
+	# We most probably want the latest point release of a specific
+	# kernel thus we add `tail -n 1` in the pipeline below.
+	#
+	local upstream_tag_info
+	upstream_tag_info=$(git ls-remote --tags --ref upstream | grep "${tag_prefix}" | tail -n 1)
+	[ -z "${upstream_tag_info}" ] && die "could not find upstream tag for tag prefix: ${tag_prefix}"
```
pzakha, line 256: nit: for consistency, it would be preferable to use `[[` everywhere

packages/linux-kernel-aws/config.sh
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# shellcheck disable=SC2034
+DEFAULT_PACKAGE_GIT_URL="https://github.com/sdimitro/dlpx-aws.git"
+DEFAULT_PACKAGE_VERSION=1.0.0
```
pzakha, line 20: `DEFAULT_PACKAGE_VERSION` is not needed here (as a matter of fact, I'm planning to get rid of this variable entirely in the future).

sdimitro:
@pzakha this should be good for another round of review.

I fixed the minor nits, updated all the repo links with the new Delphix repos, and took out any barriers that I had put for `merge_with_upstream`.

When all is good I'm squash everything in to a single commit so we can merge it.

PR #102, opened 2020-06-24
Route zfs package into recovery environment build

pcd1193182:
This PR moves the recovery-environment package into the kernel build chain so that it can take the zfs build products to install them into the recovery environment.  This PR is intended to merged in conjunction with https://github.com/delphix/recovery-environment/pull/3.  It has been tested by running ab-pre-push , and by loading the built recovery environment onto a test VM and verifying that the system can import pools while in the recovery environment.

### DIFF
--- a/packages/recovery-environment/config.sh
+++ b/packages/recovery-environment/config.sh
@@ -19,8 +19,15 @@
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/recovery-environment.git"
 DEFAULT_PACKAGE_VERSION=1.0.0
 
+ZFS_DEB_PATH="$TOP/packages/zfs/tmp/artifacts"
 function prepare() {
+	if [ ! "$(ls -A "$ZFS_DEB_PATH")" ]; then
+		logmust "$TOP/buildpkg.sh" zfs
+	fi
+
 	logmust install_build_deps_from_control_file
+	logmust mkdir "$WORKDIR/repo/external-debs/"
+	logmust cp "$ZFS_DEB_PATH/"*.deb "$WORKDIR/repo/external-debs/"
 	return
 }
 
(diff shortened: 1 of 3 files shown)


packages/recovery-environment/config.sh
```diff
@@ -19,8 +19,15 @@
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/recovery-environment.git"
 DEFAULT_PACKAGE_VERSION=1.0.0
 
+ZFS_DEB_PATH="$TOP/packages/zfs/tmp/artifacts"
 function prepare() {
+	if [ ! "$(ls -A $ZFS_DEB_PATH)" ]; then
```
jwk404, line 24: Curious why you chose `ls -A` over `[ ! -d ...]` here?

pcd1193182: -d checks if the path exists and is a directory; it doesn't check if the directory is empty or not, which is what this command does.

packages/recovery-environment/config.sh
```diff
@@ -19,8 +19,15 @@
...
+		logmust "$TOP/buildpkg.sh" zfs
+	fi
+
 	logmust install_build_deps_from_control_file
+	logmust mkdir "$WORKDIR/repo/external-debs/"
+	logmust rsync -a "$ZFS_DEB_PATH/" "$WORKDIR/repo/external-debs/"
```
pzakha, line 30: I think we can limit this to `.deb` files only as there's a bunch of other unrelated files in the artifacts

pcd1193182: I tried that, and for some reason it didn't work. Originally this was `cp -r "$ZFS_DEB_PATH/*.deb" "$WORKDIR/repo/external-debs/"` but it consistently failed with "No such file or directory" and I wasn't able to determine why. `ls`ing the path revealed plenty of files matching the blob.

pzakha: Ah, that's because you quoted the * so it takes it literally. Try this:
```
cp -r "$ZFS_DEB_PATH/"*.deb "$WORKDIR/repo/external-debs/"
```

PR #123, opened 2020-09-16
TOOL-10145 linux-pkg rework: main linux-pkg change

pzakha:
This is the main linux-pkg commit in the linux-pkg rework project (TOOL-9778).

**See [design document](https://docs.google.com/document/d/1wD5VZGhTLac7xzKZSJyKkJq72ailorCivJwppP0tkWQ) for more information.** 

I've split this PR into 5 commits to make it easier for reviewing, but those commits are not independent from one another, so they will be pushed together.

Here's a summary of each commit:
1. Each package now publishes artifacts to its own S3 directory. linux-pkg packages that have build dependencies on other linux-pkg packages fetch those dependencies from S3.

2. There are 5 new packages for each linux kernel flavour. The linux kernel packages are different from regular packages because they can be generated from 3 different sources, each way having its own config file. While I've tested that packages can be "built" from all of the 3 sources, for now (on master) we only use the "archive" source. The "archive" source essentially means that we are apt-download'ing debs that are provided by Ubuntu from the ubuntu archive (or rather from our local mirror of the archive: the linux-package-mirror).

3. buildpkg would previously try to do many things at once and the logic was becoming confusing. Previously it would be able to build a package, sync a package with upstream, and generate the initial commit for a new package. With the upcoming change buildpkg will only be responsible of building a package.

4. The logic to sync a package with upstream was also modified. There are now 3 separate commands:
    - `checkupdates.sh <package>`: Check if a package has updates from upstream.
    - `sync-with-upstream.sh <package>`: This function has two responsibilities. Let's say for the following example that we are running for the "master" product branch. First it will pull the latest changes from upstream and push them to our repository's "upstreams/master" branch. Second it will attempt to merge the changes from "upstreams/master" into "master", and push the changes to a staging branch for building and testing.
    - `push-merge.sh <package>`: This last command is intended to be called once building and testing of the package is complete. It will push the staged changes generated by `sync-with-upstream` to the "master" branch.

5. Remove all the logic that is now obsolete. This includes: 
    - the buildlist / updatelist commands, since now each command targets a single package and Jenkins takes care of figuring out which packages must be build/updates via the `query-packages.sh` command.
    - the build-info package. Given that we do not have a buildlist anymore, the build-info packages are no longer generated.
    - some of the old package lists, which were replaced by new ones.
    - the push-updates.sh command that has been replaced by new logic in commit 4.

## Dependencies
See [work plan](https://docs.google.com/document/d/1wD5VZGhTLac7xzKZSJyKkJq72ailorCivJwppP0tkWQ/edit#heading=h.st5bs7ygy0q6)

## Testing
See [test worksheet](https://docs.google.com/spreadsheets/d/1PXy6vk92tnhI9B1FHTrR1Ay7YR2_qVRyga46WYKfwwQ)


packages/linux-kernel-oracle/config.delphix.sh
```diff
@@ -0,0 +1,43 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# shellcheck disable=SC2034
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/linux-kernel-oracle.git"
+
+UPSTREAM_GIT_URL="https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-oracle/+git/bionic"
+UPSTREAM_GIT_BRANCH="@PLACEHOLDER-WORKAROUND@"
```
sdimitro, line 22: what are your thoughts about the future for this? lgtm for this review btw

packages/zfs/config.sh
```diff
@@ -163,9 +163,6 @@ function build() {
 	done
 	logmust cd "$WORKDIR"
 	logmust mv "all-packages/"*.deb "artifacts/"
-
-	# Install libzfs which is required to build grub
-	logmust install_pkgs "$WORKDIR/artifacts"/{libnvpair1linux,libuutil1linux,libzfs2linux,libzpool2linux,libzfslinux-dev}_*.deb
```
sdimitro, line 168: It's nice seeing these dependencies going one way instead of being declared everywhere :)

sdimitro (approved):
A couple miscellaneous questions - overall LGTM, appreciative of the source comments too

packages/linux-kernel-oracle/config.delphix.sh
pzakha: I haven't given it much thought to be honest. We could remove the requirement to have UPSTREAM_GIT_BRANCH present when UPSTREAM_GIT_URL is there, but I find it's a useful check especially for folks who are less familiar with linux-pkg. Perhaps we could use "none" here to be more consistent with DEFAULT_PACKAGE_GIT_URL.

packages/adoptopenjdk/config.sh
```diff
@@ -23,10 +23,7 @@ tarfile="OpenJDK8U-jdk_x64_linux_hotspot_8u262b10.tar.gz"
 jdk_path="/usr/lib/jvm/adoptopenjdk-java8-jdk-amd64"
 
 function prepare() {
-	if ! ls "$TOP/packages/make-jpkg/tmp/artifacts/"*deb >/dev/null 2>&1; then
-		echo_bold "custom java-package not installed. Building package 'make-jpkg' first."
-		logmust "$TOP/buildpkg.sh" make-jpkg
-	fi
+	logmust install_pkgs "$DEPDIR"/make-jpkg/*.deb
```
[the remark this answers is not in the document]
pzakha: Yes, so for each dependency we only need a certain package to be installed. We could perhaps add something like:
```
install_dep_pkgs <dep> <pkg1> <pkg2> ...
```

That said there's a few exceptions where we'd still want to resort to globs, like `install_kernel_headers()` and delphix-sso-app/config.sh which does `logmust install_pkgs "$DEPDIR"/adoptopenjdk/*.deb`.

pzakha:
@sdimitro I've posted some fixes and I've also updated the README.

sebroy (approved):

PR #127, opened 2020-11-11
TOOL-10581 linux-pkg should be able to use git token to push to remote

pzakha:
This now allows the auto-update logic to push changes to the remote using a git token. Previously only a combination of a user & password were supported.

GitHub is deprecating the usage of user:password auth for some of its APIs (https://developer.github.com/changes/2020-02-14-deprecating-password-auth/). It's unclear if `git push` is affected by this, but in any case using a token is a much better security practice.

## Next Steps
Update Jenkins to pass a token instead of a user & password (see TOOL-10580).

## Testing
Tested by running `./sync-with-upstream.sh bpftrace` manually on a bootstrap VM, targetting my developer fork of the `bpftrace` repo.
- Tested passing PUSH_GIT_TOKEN in env
- Tested passing PUSH_GIT_USER & PUSH_GIT_PASSWORD in env
- Tested not passing anything in env and getting prompted for a token

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -857,19 +857,17 @@ function merge_with_upstream_default() {
 # provided in env.
 #
 function check_git_credentials_set() {
-	if [[ -z "$PUSH_GIT_USER" ]] || [[ -z "$PUSH_GIT_PASSWORD" ]]; then
+	if [[ -z "$PUSH_GIT_TOKEN" ]] && [[ -z "$PUSH_GIT_USER" || -z "$PUSH_GIT_PASSWORD" ]]; then
 		if [[ -t 1 ]]; then
 			if [[ "$DRYRUN" == "false" ]]; then
 				echo_bold "WARNING: this is NOT a dry-run, you are pushing to" \
 					"a production branch"
 			fi
-			echo "Please enter git credentials to push to remote ($DEFAULT_PACKAGE_GIT_URL)."
-			read -r -p "Username: " PUSH_GIT_USER
-			read -r -s -p "Password: " PUSH_GIT_PASSWORD
-			export PUSH_GIT_USER
-			export PUSH_GIT_PASSWORD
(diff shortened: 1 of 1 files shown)


pzakha:
cc: @grodr

sdimitro (approved):

lib/common.sh
```diff
@@ -857,19 +857,17 @@ function merge_with_upstream_default() {
...
-	if [[ -z "$PUSH_GIT_USER" ]] || [[ -z "$PUSH_GIT_PASSWORD" ]]; then
+	if [[ -z "$PUSH_GIT_TOKEN" ]] && [[ -z "$PUSH_GIT_USER" || -z "$PUSH_GIT_PASSWORD" ]]; then
 		if [[ -t 1 ]]; then
 			if [[ "$DRYRUN" == "false" ]]; then
 				echo_bold "WARNING: this is NOT a dry-run, you are pushing to" \
 					"a production branch"
 			fi
-			echo "Please enter git credentials to push to remote ($DEFAULT_PACKAGE_GIT_URL)."
-			read -r -p "Username: " PUSH_GIT_USER
-			read -r -s -p "Password: " PUSH_GIT_PASSWORD
```
grodr, line 868: Will this section still be needed in the case that PUSH_GIT_USER and PUSH_GIT_PASSWORD are passed in?

grodr (approved):
Other than than the one comment I left, LGTM! Ship it!

lib/common.sh
pzakha: This whole section (`if [[ -t 1 ]]; then`) is only for when we run the scripts manually, in which case none of the PUSH_* variables will usually be passed.

grodr: Ah, okay, makes sense.

sebroy (approved):

PR #130, opened 2020-12-03
DLPX-73310 linux-pkg: enable building from release tag

pzakha:
When using a release tag in DEFAULT_GIT_BRANCH the linux-package-mirror
link and the base url for fetching build dependencies are now properly
set up.

## Testing
- ab-pre-push (rebuilding zfs + dependenciees & targetcli-fb): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4395
- manually tested that the logic works on bootstrap VM. Note that given that we do not currently have a release which uses "combined-packages" artifacts, I had to provide a custom `DEPENDENCIES_BASE_URL` link to make it work; that said I verified that the logic from `determine_dependencies_base_url()` should be compatible with artifacts that will be generated for 6.0.6.0 and later.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -26,7 +26,6 @@ export SUPPORTED_KERNEL_FLAVORS="generic aws gcp azure oracle"
 # for testing purposes to use jenkins-ops.<developer> instead.
 #
 export JENKINS_OPS_DIR="${JENKINS_OPS_DIR:-jenkins-ops}"
-export _BASE_S3_URL="s3://snapshot-de-images/builds/${JENKINS_OPS_DIR}/devops-gate/master"
 
 export UBUNTU_DISTRIBUTION="bionic"
 
@@ -142,6 +141,11 @@ function determine_default_git_branch() {
 		"$DEFAULT_GIT_BRANCH set in branch.config."
 
 	export DEFAULT_GIT_BRANCH
+}
+
+function is_release_branch() {
+	check_env DEFAULT_GIT_BRANCH
+	[[ "$DEFAULT_GIT_BRANCH" == release/* ]]
 }
 
 function check_package_exists() {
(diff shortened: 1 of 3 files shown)


lib/common.sh
```diff
@@ -144,6 +143,11 @@ function determine_default_git_branch() {
 	export DEFAULT_GIT_BRANCH
 }
 
+function is_release_branch() {
+	check_env DEFAULT_GIT_BRANCH
+	[[ "$DEFAULT_GIT_BRANCH" == release/* ]]
```
prakashsurya, line 148: I haven't looked at all the changes, but when we eventually support building branches based off a tag, rather than the tag itself.. e.g. hotfixes, which may be a series of commits branched from a release tag.. will we update this function to account for that case (somehow)?

prakashsurya: I don't mean to imply we need to handle that case right now, just want to understand the direction we're moving in, since I think we'll need to (eventually) support building from the release tag, as well as a branch (and/or detached series of commits) based off a release tag.

pzakha: So `DEFAULT_GIT_BRANCH` is mostly to set-up the defaults that would apply to:
1. All packages (sources & dependencies)
2. The linux-package-mirror link

My understanding is that we would only have a hotfix tag for packages that are modified for that hotfix, whereas for the rest we would use the defaults of the release tag that the hotfix is built for. My assumption is also that by default we would also not have a separate linux-package-mirror link for the hotfix (although we could provide one if we wanted to).

The way I was envisioning rebuilding packages for a hotfix is that the `build-hotfix` job would identify which packages are targetted by the hotfix and then pass them to the `build-packages` job via `PACKAGE_GIT_BRANCHES` (alternatively we could pass the hotfix number to build-packages and let it figure out which packages need to be rebuilt).

### Here's an example to illustrate this
Let's say we are building hotfix HF-5678, which targets version 6.1.1.0 and packages zfs & bpftrace. build-packages will set `PACKAGE_GIT_BRANCHES` to `release/6.1.1.0` and run the build-package job for zfs and for bpftrace by passing `PACKAGE_GIT_BRANCH` set to `HF-5678`.

The build-package job for zfs will then checkout linux-pkg on tag `release/6.1.1.0` and run `./setup.sh` which will set the linux-package-mirror to `http://linux-package-mirror.delphix.com/release/6.1.1.0`, and then run `./buildpkg.sh -b HF-5678 zfs`. buildpkg will fetch build dependencies for zfs from the default S3 location for the 6.1.1.0 artifacts.

prakashsurya: OK, I think I understand better... so, `DEFAULT_GIT_BRANCH` is used (at least one of it's uses) sort of how we use `UPSTREAM_PRODUCT_BRANCH` in some of our other Jenkins jobs? and thus, for a hotfix, it'll always refer to the release tag that the hotfix is branched from; is that right?

pzakha: Yeah that's right, DEFAULT_GIT_BRANCH is actually set to the value of UPSTREAM_PRODUCT_BRANCH. We could have updated that in linux-pkg for consistency with the other projects but never did. And for the hotfix question, that is correct.

prakashsurya (approved):

sebroy (approved):

PR #132, opened 2021-01-07
DLPX-73610 pin unzip package to version 6.0-21ubuntu1

pzakha:
This is a temporary workaround to DLPX-73555. It should allow resuming the sync of the linux-package-mirror on master and 6.0/stage while we are working on fixing DLPX-73555.

## Testing
Build new package: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg/job/custom/job/build-package/job/custom/1/console
Run combine-packages: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg/job/master/job/combine-packages/job/pre-push/171
Run ab-pre-push: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4568/


packages/unzip-prebuilt/config.sh
```diff
@@ -0,0 +1,37 @@
...
+# This downloads version 6.0-21ubuntu1 of the unzip package instead of the
+# latest version (6.0-21ubuntu1.1 as of writing this) which causes
+# issues due to DLPX-73555.
+#
+function fetch() {
+	logmust cd "$WORKDIR/artifacts"
+	local package="unzip_6.0-21ubuntu1_amd64.deb"
+	local url="http://artifactory.delphix.com/artifactory"
+
+	logmust wget -nv "$url/linux-pkg/unzip-prebuilt/$package" -O "$package"
```
prakashsurya, line 31: why do we have to use a prebuilt package, rather than building from source like we normally do?

pzakha: The package comes from Ubuntu directly, I've just `apt-get download`ed it into artifactory. Maybe the "prebuilt" term is confusing? I used that term because we used it for a few other packages where we only fetch the debs (like linux-prebuilt, td-agent-prebuilt).

prakashsurya: My question still stands.. why aren't we building this from source like we normally do?

pzakha: We build from source when we make modifications to the source code. In this case I don't see the need of creating a new repository on github, which is significantly more heavy-weight than just using an existing package.

prakashsurya: In general, I don't like the idea of using prebuilt binaries/packages; especially when we have no automated process to regenerate those prebuilt binaries. In this case, if we lost the prebuilt binary in artifactory, and could no longer pull it from Ubuntu, how would we recover?

Further, I also don't agree with the notion that we should only build from source when we need to make changes. I also don't agree with the notion that we should take the solution that's quickest/easiest to implement.

Without a good reason not to treat this package the same as our other linux-pkg built packages, I'm not sure I can approve this design.

If it's too cumbersome to use linux-pkg how it's intended to be used (e.g. with a new git repository, and building from that repository), then we should improve that process, rather than skirt around the problem by using manually generated (i.e. manually copied into artifactory, with no vetting process, no documentation, and not necessarily reproducible) binaries and/or packages.

Feel free to land, though, if you get approval from others; I won't block this, if this is the build architecture the rest of the team wishes to take.

pzakha: > In this case, if we lost the prebuilt binary in artifactory, and could no longer pull it from Ubuntu, how would we recover?

That's a valid concern, although we'd also have the deb permanently stored in S3 for whatever release it ends up in. 

As for your other concerns I can see that you might have an issue with a manually built package (although when you raised those concerns in the past I did provide documentation and ways to reproduce the build), however this is literally just downloading a deb with a specific version.

Linux-pkg is designed to be able to handle already built packages, and in this case I'm leveraging this capability, so I'm not sure what makes you so uncomfortable about this.

prakashsurya: > however this is literally just downloading a deb with a specific version.

To me, the method to produce the artifact is not relevant; I view both equally intractable.

> Linux-pkg is designed to be able to handle already built packages, and in this case I'm leveraging this capability,

Just because it's possible, doesn't mean that it should be encouraged.

> so I'm not sure what makes you so uncomfortable about this.

Can you clarify when we should use artifactory + prebuilt artifacts, vs. when we should use a git repository and build from source? As is, it's really not clear to me.

As an example, what approach would we take to deliver a hotfix (with the new hotfix project design), when the change needs to occur in a package built by Ubuntu. It sounds like we should build the `.deb` and store it in artifactory, then use linux-pkg to pull down that pre-built `.deb` such that it winds up in the hotfix image?

Further, if Ubuntu already had a fixed package that we wanted to deliver as a hotfix, we'd do the same `apt-get download` + artifactory + linux-pkg, as you're doing here?

I don't see those case(s) much different than what you're proposing in this review, and I want to fully understand when and why to use which approach (i.e. when and why to build from source, and when not to).

pzakha: > Can you clarify when we should use artifactory + prebuilt artifacts, vs. when we should use a git repository and build from source? As is, it's really not clear to me.

In my opinion, we should use a git repository when we need to make code changes, and avoid using a git repository otherwise, as it adds additional maintenance burden (not so much from the linux-pkg side as from the github and release maintenance side). We can discuss whether or not artifactory is the best place to store the prebuilt artifacts (and if so then what kind of directory structure we should use), but in my opinion we should avoid creating extra Github repositories when that's not necessary.

> As an example, what approach would we take to deliver a hotfix (with the new hotfix project design), when the change needs to occur in a package built by Ubuntu. It sounds like we should build the .deb and store it in artifactory, then use linux-pkg to pull down that pre-built .deb such that it winds up in the hotfix image?

We can create a github fork for the package and add it to linux-pkg the same way we do for any other package outside of the hotfix workflow. We'll just need to tag the changes with the hotfix tag.

> Further, if Ubuntu already had a fixed package that we wanted to deliver as a hotfix, we'd do the same apt-get download + artifactory + linux-pkg, as you're doing here?

In that case, yeah we could do exactly what we are doing in this review; the hotfix workflow shouldn't be any different from a regular product workflow.

Feel free to schedule a meeting if you want to discuss this more.

prakashsurya: > as it adds additional maintenance burden (not so much from the linux-pkg side as from the github and release maintenance side).

IMO, it's relatively trivial to create a git repository in gitlab. I don't see any requirement to use github with linux-pkg.

Additionally, if we don't have the same release branch requirements for the prebuilt artifact, I don't see why we need to have the same release branch requirements when building from the git repository; thus reducing the "maintenance burden" you're referring to.. i.e. we can always just build and use the "master" branch, no matter the release.. just like we use the same prebuilt package, no matter the release.

> Feel free to schedule a meeting if you want to discuss this more.

I don't want to discuss any more, but I'd like some documentation we can use to guide folks to do the right things based on these decision you've made; e.g. I don't see any mention of this "prebuilt" workflow in the README.

prakashsurya:
Out of curiosity, if we're pulling in a prebuilt package from artifactory, what is the benefit of doing this in linux-pkg, vs. doing this in appliance-build? e.g. perhaps in the "build-ancillary-repository" step?

pzakha:
> Out of curiosity, if we're pulling in a prebuilt package from artifactory, what is the benefit of doing this in linux-pkg, vs. doing this in appliance-build? e.g. perhaps in the "build-ancillary-repository" step?

Not much of a benefit to be honest, I think we could very well do it from there. I can take a look at that.

prakashsurya:
I don't mean to imply I'd rather us do this in appliance-build, I'm just trying to fully understand the underlying architecture for the design of our build system.. and why we do what we do, where we do it.

packages/misc-debs/config.sh
```diff
@@ -0,0 +1,81 @@
...
+	#   now fails to build due to broken third party dependencies. See
+	#   DLPX-69338 and DLPX-68211.
+	# - unzip was added as a temporary workaround to DLPX-73555.
+	#
+	local debs=(
+		"td-agent_3.5.0-delphix-2019.09.18.20_amd64.deb 84dfa2436039ff2a6312484bd7295ebaf570b5f59f100380b57e68b4800855c4"
+		"unzip_6.0-21ubuntu1_amd64.deb d46069c369ce88c8dd91c52abb8de8d6053606748ef18b3b9bc290fdd8ad2953"
+	)
+
+	local url="http://artifactory.delphix.com/artifactory/linux-pkg/misc-debs"
```
prakashsurya, line 62: can we embed this URL in the `fetch_file_from_artifactory` function? if that function is specific to artifactory, as it's name suggests, I feel like we could just hardcode the URL for artifactory into that function, rather than have the callers know the correct URL to use. what do you think?

prakashsurya (approved):

pzakha:
After having an offline chat with Prakash, I've made a few changes to how we fetch already built debs.

Here's a summary of the changes:
1. This is adding a new meta linux-pkg package responsible of fetching all the already-built debs we want to use. Having a single meta package will make it easier to add/remove debs from there in the future.
2. I've added instructions for how and in which cases to add additional debs.
3. I've added a SHA256 check to fetched packages to make sure that the binaries downloaded are what we expect.

Additional changes:
- removed the `java8` package since it hasn't been used for a long time.
- moved td-agent-prebuilt package into this new misc-debs package.

packages/misc-debs/config.sh
pzakha: I've thought of that, however we sometimes fetch from other artifactory buckets (see for instance the `adoptopenjdk` package). We could provide whatever is after "http://artifactory.delphix.com/artifactory/", however I figured it's more clear if we just pass the whole URL, rather than the "path under http://artifactory.delphix.com/artifactory/".

sebroy (approved):

PR #137, opened 2021-02-12
CP-4543 Write package to pull down Base Docker image of python:2.7.18-slim and save it to disk

nhlien93:
The goal of this package is just to leverage docker to pull down a specific docker image (python:2.7.18-slim) and then save it as a docker file (again using docker). This file is then placed /opt/delphix/server/etc/docker-python-2-7-18-slim.tar 
Manually tested that after this package is built, it was able to pull down as a dependency of the delphix-virtualization package and we can see now that the tar file is located in the right location.

I do think I'd like to somehow make the line in the rules file shorter would be nice so if someone knows the syntax would love the help.

This is a feature necessary for the Plugin Runtime project where our goal is to have the base python image available to be loaded into the docker daemon on stack startup.

http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/4826/

Verified by cloning: dlpx-lnguyen-master, ami-0b3beb0df43c6f2fb


pzakha (approved):
LGTM, can you add a link to your appliance-build-orchestrator testing?

Also, something I forgot to mention, after you've integrated this you should manually run http://ops.jenkins.delphix.com/job/devops-gate/job/master/job/generate-jobs/ so that it can create a Jenkins job for the new package, otherwise the build will be failing for some time. 

Alternatively, you can first push a change without adding `docker-python-image` to `package-lists/build/main.pkgs`, wait for generate-jobs to run (or run it manually yourself), then submit a new change to add `docker-python-image` to the list once the new package has been built by the automation.

pzakha (approved):

packages/docker-python-image/code/debian/rules
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+%:
+	dh $@
+
+override_dh_install:
+	mkdir -p debian/tmp/opt/delphix/server/etc
```
sebroy, line 22: `/opt/delphix/server/etc` contains configuration files and metadata. A `/lib` directory might be more appropriate?

sebroy (changes requested):
This package seems quite unique from all other packages in that all package files are being maintained in the `linux-pkg` repo itself rather than in a separate repo. Can you comment on the constraints and tradeoffs that led to this choice?

pzakha:
> This package seems quite unique from all other packages in that all package files are being maintained in the linux-pkg repo itself rather than in a separate repo. Can you comment on the constraints and tradeoffs that led to this choice?

I can comment on that since I'm the one who suggested Lindsay to take that route. Here are the trade-offs:

Pros:
- Less friction to setup since we do not need to create and configure a separate repository
- One less repository to maintain

Cons:
- Building the new package will not trigger automatically when the package is updated
- Package will be rebuilt on every release regardless of whether it has changes or not

The advantage with the current approach is that we can easily switch to a standalone repository if we want to go that route. All we need to do is to copy the contents of the `code` directory into a new repository, set `DEFAULT_PACKAGE_GIT_URL` and delete the custom `fetch()` hook.

packages/docker-python-image/config.sh
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL=none
+DEFAULT_PACKAGE_VERSION="1.0.0"
```
sebroy, line 20: Would it make sense to mirror the versioning of the underlying docker image?

packages/docker-python-image/code/debian/rules
nhlien93: Honestly I don't really know what directory is used for what but I knew there was custom files in etc for different data platforms so that's why I went with that.

Do you think /lib should be the actual location it should go in or would you recommend creating a new directory inside it?

If you have any suggestions please let me know as I was definitely not sure what location was ideal for a file like this (something that needs to be used in the app gate code).

sebroy:
Regarding this pro:
> Gives a good example for developers that are new to our package build workflow that want to add a new package

I'm not sure I see it the same way. Given that this is not how our package build workflow generally works, this would be a "not good" example (right?)...

pzakha:
> I'm not sure I see it the same way. Given that this is not how our package build workflow generally works, this would be a "not good" example (right?)...
I guess, if we had a simple separate repository that should work too
I'm personally fine either way. If you think a separate repository would be better I'd not oppose it.

packages/docker-python-image/config.sh
nhlien93: I think that could work? We don't have plans to be downloading different versions of this docker image however. Eventually we may be using 3.0 (which then I may have to create a new version of this package.)

Feels slightly weird to have the first package be versioned 2.7.18 and then the second version of this package be 3.x.x (whenever we switch to python 3)?

But I honestly don't know what these versions are used for or how important they are so definitely am willing to change it to whatever people think makes the most sense.

nhlien93:
> I guess, if we had a simple separate repository that should work too
I'm personally fine either way. If you think a separate repository would be better I'd not oppose it.

I'm also okay to do whatever! Just let me know. I did feel like creating a full new repo for a couple files felt a little bit excessive however if this is the typical workflow to use linux-pkg and we want others to do the same in the future I can definitely make that happen.

packages/docker-python-image/config.sh
pzakha: We haven't really been using the version fields much for in-house packages. Those could be leveraged if we wanted to apply some conditional upgrade logic based on a package's previous version, but so far we haven't needed that.

As a side note, I'm planning to get rid of `DEFAULT_PACKAGE_VERSION` some time in the future, and move that information into the package's `debian/copyright` file like it's normally done for packages.

packages/docker-python-image/code/debian/rules
sebroy: I think `/opt/delphix/server/lib` would work nicely (this is where we install jar's and other binary package deliverables).

packages/docker-python-image/config.sh
sebroy: I see, and that's ok because the build portion of the version does change, enabling the upgrade of that package when it gets updated? If that's the case, then dropping this comment is fine, I was just curious about whether this mattered.

pzakha: Yeah, the final, full version of the package is done by concatenating the package's version, determined by DEFAULT_PACKAGE_VERSION, and the package revision, which is auto-generated from a timestamp. For third party packages we use various methods to determine the version, but instead we could just automatically extract it from the `debian/changelog` file. We should be able to do this for all the packages.

sebroy (approved):

PR #139, opened 2021-02-25
TOOL-11187 leverage linux-pkg to build delphix-masking package

prakashsurya:


package-lists/build/main.pkgs
```diff
@@ -10,6 +10,7 @@ cloud-init
 crash
 crash-python
 crypt-blowfish
+delphix-masking
```
pzakha, line 13: we are currently using `masking` instead of `delphix-masking`, so it's probably better to keep it that way if we want to minimize changes.

Also, we would probably want to add the new package to `main.pkg` in a follow-up PR, as from that point on the combine-packages job will expect to find masking package's artifacts in `s3://.../linux-pkg/master/build-package/masking/post-push/latest`. It would also likely conflict with the current special case logic that we use in the `linux_pkg_*` jobs for building the masking package.

pzakha (commented):
The logic looks reasonable. I think it would be useful here to identify other repositories that will need to be changed for this move, sync up with the masking team regarding this change, and establish a deployment plan to make sure to minimize disruptions.

package-lists/build/main.pkgs
prakashsurya: > we are currently using masking instead of delphix-masking

what do you mean? this name is only for linux-pkg, so what do you mean we're currently using "masking"?

do you mean, the debian package is named "masking"? if so, I don't think I'm changing that, rather I'm naming the linux-pkg concept to "delphix-masking" (to match some of our other delphix names), while the debian package name should remain the same as before.

prakashsurya: looking a little closer, it appears the package name is "delphix-masking":
```
520     packageName 'delphix-masking'
```
so I think it'd be best to be consistent with that, and use delphix-masking for the linux-pkg name for this package.

pzakha: We could use `delphix-masking`, we'd just need to make changes in `appliance-build-stage0` and `git-utils` that will use the right name depending on the version being built

packages/masking/config.sh
```diff
@@ -0,0 +1,50 @@
...
+	JAVA_HOME=$(cat "$DEPDIR/adoptopenjdk/JDK_PATH") ||
+		die "Failed to read $DEPDIR/adoptopenjdk/JDK_PATH"
+
+	logmust cd "$WORKDIR/repo"
+
+	logmust jq -n \
+		--arg h $(git rev-parse HEAD) \
+		--arg d $(date --utc --iso-8601=seconds) \
+		'{ "dms-core-gate" : { "git-hash" : $h, "date": $d }}' \
+		>"$WORKDIR/artifacts/metadata.json"
```
pzakha, line 37: Can you add a comment for what is consuming this data?

pzakha (approved):
Do you have a deployment plan for how you will deploy this changes along with the devops-gate changes to minimize disruptions?

prakashsurya:
> Do you have a deployment plan for how you will deploy this changes along with the devops-gate changes to minimize disruptions?

My plans are to: 1. land this change, 2. build the masking package via linux-pkg, then 3. push the related devops-gate changes and add this package to `main.pkgs`.

packages/masking/config.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+#
+# Copyright 2020 Delphix
```
pcd1193182, line 3: 2021

pcd1193182 (approved):

PR #149, opened 2021-03-23
TOOL-11245 leverage linux-pkg to build delphix-virtualization package (#145)

prakashsurya:


lib/common.sh
```diff
@@ -470,7 +470,7 @@ function install_pkgs() {
 		echo "Running: sudo env DEBIAN_FRONTEND=noninteractive " \
 			"apt-get install -y $*"
 		sudo env DEBIAN_FRONTEND=noninteractive apt-get install \
-			-y "$@" && return
+			-y --allow-downgrades "$@" && return
```
sdimitro, line 473: 2 small things here:
[1] Would it be possible to add a comment on why we added `--allow-downgrades` ? Besides adding the why, can you provide the specific package names as examples too?
[2] I think we should also update the `echo` command right before this statement in case someone is looking at the logs and gets confused.

sdimitro (approved):
Besides one nit and a question, the code LGTM,

I assume you tested this with Jenkins?

lib/common.sh
prakashsurya: Should I address these comments in this backport? Or open a follow on PR to master? I'm leaning to a master-based PR, is that reasonable to yoy?

prakashsurya: I've opened #150

prakashsurya:
> I assume you tested this with Jenkins?

I haven't specifically on 6.0/stage, since our policy generally has been clean backports don't require re-verification; but I can test this and report back.

prakashsurya:
`build-packages` is [here](http://collins.d.delphix.com:29627/job/devops-gate/job/master/job/linux-pkg/job/6.0/job/stage/job/build-packages/job/post-push/8/)
`build-package` is [here](http://collins.d.delphix.com:29627/job/devops-gate/job/master/job/linux-pkg/job/6.0/job/stage/job/build-package/job/virtualization/job/post-push/1/)

matt-skinner-delphix (approved):

PR #150, opened 2021-03-23
Add comment to explain use of "--allow-downgrades"

prakashsurya:


sdimitro (approved):
That makes sense. Thank you for taking the time to explain this!

sebroy (approved):

PR #151, opened 2021-04-13
CP-4898 Write gradle task to download grpcio and generate grpc files

nhlien93:
This change adds virtualenv to the build dependencies when the virtualization package is being built. This is needed so that some pip and python commands can be run to generate files for the plugin runtime project.

Confirmed from a run that linked linux-pkg that this was built properly and virtualenv was used to install something via gradle during the build.

This diff is needed for http://reviews.delphix.com/r/67972/ to be pushed.


rasantel (approved):

jeffngo (approved):
LGTM

PR #156, opened 2021-05-05
DLPX-74479 Create a separate host-jdks package for all files under /u…

aditya-joshi-delphix:
…sr/share/host-jdks

# Context:
This change is part of CP-4190 (Self-service hotfix and automation - Phase I)
We want to separate host-jdks which are currently part of delphix-virtualization debian package into a separate debian package to reduce size of the hotfix build. 

# Problem:

Currently the size of all host jdk files under /opt/delphix/host is close to 1GB. We don't want to build JDKs every time a new hotfix build is generated. Thus separating these files into a separate package seems logical. 

delphix@ip-10-110-202-68:~$ sudo find /opt/delphix/host -type f -size +10M -exec du -h {} \; | sort -n
45M	/opt/delphix/host/windows/DelphixConnectorInstaller.exe
75M	/opt/delphix/host/jdk/linux_x86/jdk1.8_172/jdk.tar.gz
96M	/opt/delphix/host/jdk/linux_ppc64le/jdk1.8/jdk.tar.gz
96M	/opt/delphix/host/jdk/windows_x86/jdk1.8/jdk.zip
99M	/opt/delphix/host/jdk/linux_x86/jdk1.8/jdk.tar.gz
101M	/opt/delphix/host/jdk/sunos_x86/jdk1.8/jdk.tar.gz
104M	/opt/delphix/host/jdk/sunos_sparc/jdk1.8/jdk.tar.gz
115M	/opt/delphix/host/jdk/hpux_ia64/jdk1.8/jdk.tar.gz
145M	/opt/delphix/host/jdk/aix_powerpc/jdk1.8/jdk.tar.gz

# Solution:

Generate an independent host-jdks debian package. 

# Testing:

 http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg/job/custom/job/build-packages/job/custom/71/console (PASSED)

# Implementation:
Created a new package for host-jdks when unpacked the JDK files will be located at /usr/share/host-jdks

# Notes To Reviewers:
This is my first linux-pkg PR :) I am not sure about the order of how these changes should go in. I am willing to split the change into multiple reviews if needed. 

# Deployment Plan:
This change needs to go in along with the corresponding app-gate change, https://gitlab.delphix.com/aditya.joshi/dlpx-app-gate/commit/3ebf4b64af50d5145d151ab1b52484eb6bb334eb

### DIFF
--- a/packages/host-jdks/config.sh
+++ b/packages/host-jdks/config.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+# Copyright 2021 Delphix
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/host-jdks.git"
+
+function build() {
+	logmust mkdir -p "$WORKDIR/repo"
+	logmust dpkg_buildpackage_default
+}


packages/virtualization/config.sh
```diff
@@ -17,7 +17,7 @@
 # shellcheck disable=SC2034
 
 DEFAULT_PACKAGE_GIT_URL="https://gitlab.delphix.com/app/dlpx-app-gate.git"
-PACKAGE_DEPENDENCIES="adoptopenjdk crypt-blowfish misc-debs"
+PACKAGE_DEPENDENCIES="adoptopenjdk crypt-blowfish host-jdks misc-debs"
```
prakashsurya, line 20: I think we need to pull both of these changes out of this PR, and push them in a seperate PR after this one lands.

packages/virtualization/config.sh
```diff
@@ -43,6 +43,7 @@ function prepare() {
 	logmust install_pkgs \
 		"$DEPDIR"/adoptopenjdk/*.deb \
 		"$DEPDIR"/crypt-blowfish/*.deb \
+		"$DEPDIR"/host-jdks/*.deb \
```
prakashsurya, line 46: I think we need to pull both of these changes out of this PR, and push them in a seperate PR after this one lands.

prakashsurya (commented):
Looks fairly reasonable to me, lets see how @pzakha feels about this. I don't know if we have any other package that's quite like this one, so there's no good example to follow..

packages/host-jdks/code/debian/copyright
```diff
@@ -0,0 +1,18 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Files: *
+Copyright: 2021 Delphix
+License: Apache-2.0
```
sebroy, line 4: Are these JDKs really covered under Apache 2.0? I thought OpenJDK was GPLv2.

sebroy (commented):
Yes!!! I can't vouch for the packaging changes (I'll let others review that in detail), but I love this change overall.

package-lists/build/main.pkgs
```diff
@@ -15,6 +15,7 @@ drgn
 docker-python-image
 gdb-python
 grub2
+host-jdks
```
prakashsurya, line 18: IIRC, this also as to be landed in a seperate PR.

packages/host-jdks/code/debian/control
```diff
@@ -0,0 +1,25 @@
...
+Source: host-jdks
+Section: misc
+Priority: optional
+Maintainer: Delphix Engineering <eng@delphix.com>
+Standards-Version: 4.1.2
+
+Package: host-jdks
+Architecture: all
+Description: Downloads host jdks from artifactory and builds a debian package
+ The JDKs will be located at /opt/delphix/host
```
prakashsurya, line 25: This path is outdated.

packages/host-jdks/code/debian/control
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright 2020 Delphix
```
prakashsurya, line 2:
```suggestion
# Copyright 2021 Delphix
```

packages/host-jdks/code/debian/control
prakashsurya, line 25:
```suggestion
 The JDKs will be located at /usr/share/host-jdks
```

pzakha:
This package is fairly unique indeed, thanks for working on this.

On one hand it fetches files from artifactory, which makes it similar to the `misc-debs` package, but on the other hand it also builds its own deb, which is not done by `misc-debs`. Ideally we'd have had the business logic of the package live in a separate repository, kinda like what we ended up doing for the `docker-python-image` package in https://github.com/delphix/linux-pkg/pull/137, however that would require a bunch of additional work and to add into that package rules file logic to fetch from artifactory. Given the extra work required and the fact that we'll probably not touch this package often I'd say it's probably not worth it at this point.

Like Prakash mentioned, you'll want to move the changes to main.pkgs and to the virtualization/config.sh out of this review. The reason being that when adding a new package you need to do things in stages: 
1. You add the new package's definition (i.e. its config.sh and other relevant changes)
2. You make sure Jenkins builds artifacts for this new package
3. You add this package to the list of packages consumed by the build (i.e. main.pkgs) and add it as a dependency to other packages (i.e. virtualization/config.sh).
If you don't do that in stages, then the build will start failing as soon as you land this PR, and until step 2 has been completed. Therefore to eliminate disruptions, we do it in stages.

You will be able to land the associated dlpx-app-gate change once you've completed those 3 steps

Regarding the testing, you'll want to also run appliance-build-orchestrator testing using the packages you've built. See https://docs.delphix.com/display/EH/Testing+Changes+to+Packages#TestingChangestoPackages-Testingchangesusingpreviouslybuiltpackages for more info.

The code changes look fine to me aside from the comments already pointed out by others, but I'll wait for links to appliance-build-orchestrator before approving

prakashsurya:
> Given the extra work required and the fact that we'll probably not touch this package often I'd say it's probably not worth it at this point.

Are we sure we want to set an example of adding package "code" like this? I'm not opposed to moving the "code" into a new repository.. I'd rather we be consistent with the normal convention, than take a shortcut here just because it's easier to land.

It really shouldn't be that much more work to create a new repository, and move the build logic into that new repository. If that'd be the better way to structure this change, I'd recommend we go with that; a little extra time spent doing things the right way now, may save us time and complexity maintaining this over the long term.

pzakha:
> It really shouldn't be that much more work to create a new repository, and move the build logic into that new repository. If that'd be the better way to structure this change, I'd recommend we go with that; a little extra time spent doing things the right way now, may save us time and complexity maintaining this over the long term.
Last time it took quite a bit of extra work and coordination to get everything setup properly with a new repository. I didn't want to impose additional work for the hotfix improvement project, which is why I wasn't blocking this, so I'll let you decide on this one.

packages/host-jdks/config.sh
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL=none
+DEFAULT_PACKAGE_VERSION="1.0.0"
```
pzakha, line 12: You can get rid of DEFAULT_PACKAGE_VERSION as it is not needed anymore now that https://github.com/delphix/linux-pkg/pull/143 has been integrated

PR #163, opened 2021-05-13
Add hostchecker2 files from virtualization build

prakashsurya:


prakashsurya:
Looking at the previous logic that we used to build the virtualization package, we had this:
```
167         dir('dlpx-app-gate/appliance') {                                        
168           /*                                                                    
169            * Move all artifacts into the "distributions" directory so that they can be picked up
170            * later by the logic that checksums all artifacts and uploads them to S3.
171            */                                                                   
172           sh """                                                                
173                 mkdir -p packaging/build/distributions/hostchecker2             
174                 cp -r bin/out/common/com.delphix.common/uem/tars packaging/build/distributions/hostchecker2
175                                                                                 
176                 cp toolkit/toolkit-devkit.tar packaging/build/distributions     
177                                                                                 
178                 cp packaging/build/reports/dependency-license/* packaging/build/distributions
179                 """.stripIndent()
```

It looks like, when we shifted the build to linux-pkg, we inadvertently left out the command:
```
cp -r bin/out/common/com.delphix.common/uem/tars packaging/build/distributions/hostchecker2
```

This change add that command to the `config.sh` file of the virtualization build.

prakashsurya:
Looking a bit closer, I actually see us attempt to copy the hostchecker tar files in the `config.sh`, here:
```
 90         logmust cp -vr \                                                        
 91                 bin/out/common/com.delphix.common/uem/tars \                    
 92                 "$WORKDIR/artifacts/hostchecker2"  
 ```
 
 so this change would be redundant, and not correct.

prakashsurya:
superseded by https://github.com/delphix/linux-pkg/pull/164

PR #169, opened 2021-05-18
zfs_object_agent build requires cargo, rustc

manoj-joseph:
zfs-object-agent is written in rust. We need to install rust devtools on the bootstrap VM before we build the ZFS packages. This change accomplishes that.

Successful build: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg/job/master/job/build-package/job/zfs/job/pre-push/123/console

Note: We had done this earlier for the delphix appliance.
https://github.com/delphix/appliance-build/pull/544

### DIFF
--- a/packages/zfs/config.sh
+++ b/packages/zfs/config.sh
@@ -29,6 +29,7 @@ function prepare() {
 		autogen \
 		autotools-dev \
 		build-essential \
+		cargo \
 		debhelper \
 		devscripts \
 		dh-autoreconf \
@@ -48,8 +49,10 @@ function prepare() {
 		lsb-release \
 		lsscsi \
 		parted \
+		pkg-config \
 		po-debconf \
 		python3 \
+		rustc \
 		uuid-dev \
 		zlib1g-dev
 	logmust install_kernel_headers


ahrens (approved):

pzakha (approved):

prakashsurya (approved):
I'm fine with this, since it maintains the status quo, but I think it'd be better if we could declare the build dependencies in the ZFS package itself, and then use `install_build_deps_from_control_file` here.

prakashsurya:
As an example of what I mean, the `libkdump` repository appears to do this well.. in the `debian/control` of the `libkdump` repository we have this:
```
Build-Depends: autoconf,
               automake,
               dh-python,
               liblzo2-dev,
               libsnappy1v5,
               libsnappy-dev,
               libtool,
               pkg-config,
               python3-distutils,
               python3.6-dev,
               zlib1g-dev
```

and then in `libkdump`'s `config.sh` file, we have this:
```
 function prepare() {                                                            
         logmust install_build_deps_from_control_file                            
 }  
 ```

I'd expect the same approach could be applied to the ZFS repository, potentially eliminating these types of cross-repository dependencies.

pzakha:
> I'm fine with this, since it maintains the status quo, but I think it'd be better if we could declare the build dependencies in the ZFS package itself, and then use install_build_deps_from_control_file here.

Yeah ZFS is a bit of a weird special case because it has 2 control files: one for the userland packages and another one for the kernel ones. I think we could make it work somewhat by installing whatever is in the "main" control file first, and then adding up the extras from control.modules.in manually, or even do do something similar to `install_build_deps_from_control_file` for the control.modules.in file.

Ideally, we should also sync up our `debian` directory with the latest from Ubuntu since things have changed quite a bit. Perhaps we can do all that during our transition to 20.04.

PR #178, opened 2021-06-29
TOOL-11815 linux-pkg: kernel sync-with-upstream logic sometimes fails to find expected tags upstream

pzakha:
Canonical has 2 ways of naming their tags:
* Ubuntu-gcp-5.4.0-1046.49
* Ubuntu-gcp-5.4-5.4.0-1046.49_18.04.1

For a given kernel version and a given distribution, only one naming scheme is being used. When two distributions (such as 20.04 and 18.04) use the same major kernel version, Canonical first applies the patches to one distribution and then creates a base tag. It then backports those changes to the older distribution, and then creates a longer tag name.

For instance with the example above we can see that the base tag exists on the 20.04 remote:
```
$ git ls-remote focal-gcp | grep 1046
7870999272937a8d2e14165409d799c2876771a3 refs/tags/Ubuntu-gcp-5.4.0-1046.49
```
While on 18.04 there is the longer tag name:
```
$ git ls-remote bionic-gcp | grep 1046
0432f0aad667c47d0d45345292fd59b4b5ee0d09 refs/tags/Ubuntu-gcp-5.4-5.4.0-1046.49_18.04.1
```

This review also fixes 2 small issues:
1. On focal we use the non-hwe kernel for the "generic" platform
2. The old logic for detecting if the kernel is up-to-date is slightly wrong: it checks if the upstream head is an ancestor of the current head, but there's a corner case where this doesn't quite work, such as when the current head was based on a tag that is between the upstream head and the current head. For instance:
```
* Delphix commit 6 (HEAD)
*  ...
* Delphix commit 1
* Previous upstream
* Ubuntu commit
* ...
* Ubuntu commit
* New upstream
* ...
```

## Testing
Ran update-upstream to update all kernels on Focal

### DIFF
--- a/default-package-config.sh
+++ b/default-package-config.sh
@@ -234,20 +234,24 @@ function kernel_update_upstream() {
 	# version and the ABI num that we got above.
 	#
 	# Note that "generic" (used mainly ESX) is a special
-	# case as we are currently using the HWE kernel image.
+	# case on bionic where we are using the HWE kernel image.
 	#
 	local tag_prefix_flavour
-	if [[ "${platform}" == generic ]] &&
-		[[ "$UBUNTU_DISTRIBUTION" == bionic ]]; then
-		tag_prefix_flavour="Ubuntu-hwe"
-	elif [[ "${platform}" == aws ]] ||
-		[[ "${platform}" == azure ]] ||
-		[[ "${platform}" == gcp ]] ||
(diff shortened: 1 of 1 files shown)


sebroy (approved):

sdimitro (approved):
LGTM and I appreciate the writeout+code_comments

PR #179, opened 2021-07-15
TOOL-11892 Bulk modify of shebang lines to use /usr/bin/env

jtamagnan-delphix:
## Problem

We hardcode the shebang of our bash scripts to use `/bin/bash`. There
are two problems with this.
 - Using that as the shebang line makes these bash scripts
   non-portable in the cases where `/bin/bash` does not exist.
 - By using that as the shebang line we do not allow users to change
   what version of `bash` they use by modifying their `PATH`.

## Solution

Use whichever version of `bash` is first on their PATH by using
`/usr/bin/env bash` instead of `/bin/bash`.

## Implementation

I greped for `#!/bin/bash` and replaced that with `#!/usr/bin/env
bash`. We never passed extra flags such as `-e` to the shebang so no
futher changes are needed.

## Testing

None

### DIFF
--- a/buildpkg.sh
+++ b/buildpkg.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2018, 2021 Delphix
 #
--- a/checkupdates.sh
+++ b/checkupdates.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2020 Delphix
 #
--- a/default-package-config.sh
+++ b/default-package-config.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2018, 2020 Delphix
 #
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2018, 2020 Delphix
 #
--- a/packages/adoptopenjdk/config.sh
+++ b/packages/adoptopenjdk/config.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2018, 2020 Delphix
 #
--- a/packages/bcc/config.sh
+++ b/packages/bcc/config.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2018, 2020 Delphix
 #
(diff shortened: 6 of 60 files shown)


pzakha (approved):

prakashsurya:
Out of curiosity, what is the motivation for doing this?

> Using that as the shebang line makes these bash scripts non-portable in the cases where /bin/bash does not exist.

What about when `/usr/bin/env` doesn't exit?

> By using that as the shebang line we do not allow users to change what version of bash they use by modifying their PATH.

Is this something we actually do?

I'm not opposed to this type of change, I'm just curious why we're choosing to do this.. I don't see any additional information in the JIRA ticket.

prakashsurya:
Additionally, is there plans to try and enforce the use of `/usr/bin/env` over `/bin/bash` moving forward? Just curious if/how we'd prevent regressions.

jtamagnan-delphix:
> Out of curiosity, what is the motivation for doing this?

I guess my motivation was selfish.. On my machine:
```
└→ ls -lah /bin/
total 8.1k
drwxr-xr-x   2 root           root     4.0k 2021-07-14 17:04 .
drwxr-xr-x  18 root           root     4.0k 2021-05-27 08:52 ..
lrwxrwxrwx   1 root           root       75 2021-07-14 17:04 sh -> ...

└→ ls -lah /usr/bin/
total 8.1k
drwxr-xr-x   2 root           root     4.0k 2021-07-14 17:04 .
drwxr-xr-x   3 root           root     4.0k 2021-05-22 13:16 ..
lrwxrwxrwx   1 root           root       66 2021-07-14 17:04 env -> ...
```

I do not have `/bin/bash` and couldn't run `./verify-query-packages.sh` because of it. 

> What about when /usr/bin/env doesn't exit?

I haven't seen any system without `/usr/bin/env`, though I also haven't seen many systems without `/bin/bash`. I do think that `/usr/bin/env` is always available.

>> By using that as the shebang line we do not allow users to change what version of bash they use by modifying their PATH.

> Is this something we actually do?

It's definitely something that one could do. I can imagine a future where OSx no longer comes with `bash` since the default shell is already `ZSH`.

> Additionally, is there plans to try and enforce the use of /usr/bin/env over /bin/bash moving forward? Just curious if/how we'd prevent regressions.

This hasn't been planned yet but I imagine it could be done.

prakashsurya:
> I do not have /bin/bash and couldn't run ./verify-query-packages.sh because of it.

:+1: ok, that sounds reasonable.

> This hasn't been planned yet but I imagine it could be done.

sure, I was just wondering if this was going to be a requirement moving forward (e.g. some build automation will not work correctly with out it), in which case enforcement would be more necessary.. but sounds like it's more of a best-effort type of thing, where some manual workflows may not work due to how a developer customizes their own development environment, but the product (and automation) should continue to work regardless.

prakashsurya (approved):

PR #187, opened 2021-09-09
Add "delphix-rust" to "main.pkgs" file

prakashsurya:


package-lists/build/main.pkgs
```diff
@@ -10,6 +10,7 @@ cloud-init
 crash-python
 crypt-blowfish
 delphix-platform
+delphix-rust
```
ahrens, line 13: Which VM types is this installing on?  I see the comment
> List of non-kernel packages to be included in the Delphix Appliance.

Since this is a build tool, we technically only need it on the build images, not the image we ship to customers.  It might also be useful on the QA images so that we can rebuild ZFS on those.

I would think that we would want to replace wherever we're including the ubuntu rust packages with the delphix-rust package?

prakashsurya: > Which VM types is this installing on?

IIRC, this doesn't install the package on any VMs, this only "registers" the package with the linux-pkg automation.. so that we can get automated builds of the package, when changes are pushed to the package's repository.. etc.. @pzakha might have to clarify, in case I misunderstand..

> Since this is a build tool, we technically only need it on the build images, not the image we ship to customers. It might also be useful on the QA images so that we can rebuild ZFS on those.

Right.. that's what I'm planning to do.

> I would think that we would want to replace wherever we're including the ubuntu rust packages with the delphix-rust package?

Yup, that's also what I'm planning to do; e.g. replace the installation of `rustc` and `cargo` with this new `delphix-rust` package in appliance-build.. I haven't gotten to those changes yet, and I need to initialize this new package in the linux-pkg ecosystem (which this PR is part of).

Once I get the package built, and part of the linux-pkg ecosystem, I'll open the PRs to replace our use of `rustc` and `cargo` with `delphix-rust`.

prakashsurya: > List of non-kernel packages to be included in the Delphix Appliance.

This sentence is probably misleading, and perhaps a more accurate statement would be be something like "packages that are available to be included" in the appliance.. without a package dependency, from something like the ZFS package (or equivalent), none of the packages in the file will get pulled into the appliance.

pzakha: > IIRC, this doesn't install the package on any VMs, this only "registers" the package with the linux-pkg automation.. so that we can get automated builds of the package, when changes are pushed to the package's repository.. etc.. @pzakha might have to clarify, in case I misunderstand..

That's correct. This only ensures that the package is included in the "combined-packages" bundle, which is then used in appliance-build to create the "ancillary packages repository", which we effectively add to /etc/apt/sources when building the appliance. If we want this new package to be installed on the appliance, we then need to either add it as a dependency of another package that we already install or to install it via our appliance-build variant-specific ansible scripts.

ahrens: Great, thanks for the education!

ahrens (approved):

pzakha (approved):

PR #190, opened 2021-09-09
Use "delphix-rust" instead of Ubuntu's Rust packages

prakashsurya:


prakashsurya:
`build-package` is [here](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg/job/master/job/build-package/job/zfs/job/pre-push/199/)

pzakha (approved):

pzakha (changes requested):
Hmm I think I approved this too quickly. You'll need to add a linux-pkg package dependency in zfs/config.sh:
```
PACKAGE_DEPENDENCIES="delphix-rust" 
```
And then install the debs from `"$DEPDIR"/delphix-rust/`. You can check for other packages like bpftrace or grub2 for how it's done.

prakashsurya:
ah, right.. done.. updated the testing link too.

pzakha:
Looks like the checks are failing because we are making some assumptions about the zfs package's dependencies, so might need to update the checks.

packages/zfs/config.sh
```diff
@@ -29,8 +29,8 @@ function prepare() {
 		autogen \
 		autotools-dev \
 		build-essential \
-		cargo \
 		debhelper \
+		delphix-rust \
```
pzakha, line 33: I think you'll need to remove this line too.

pzakha (approved):

ahrens (approved):

PR #194, opened 2021-09-23
TOOL-12213 linux-pkg: run setup.sh automatically when needed

pzakha:
This is just a convenience feature to make the process simpler for developers to use linux-pkg.

## Testing
- Manually checked that setup script is run automatically the first time I run buildpkg.sh
- Manually checked that the setup script is run again after a reboot
- Sanity test to make sure build-package job still works: http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/linux-pkg/job/master/job/build-package/job/cloud-init/job/pre-push/31

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,19 +50,12 @@ This quick tutorial shows how to build the packages managed by this framework.
 You need a system that meets the requirements above. For Delphix developers, you
 should clone the `bootstrap-18-04` group on DCoA.
 
-### Step 2. Clone this repository and run the setup script
+### Step 2. Clone this repository
 
 Clone this repository on the build VM.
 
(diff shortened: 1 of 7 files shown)


sdimitro (approved):
LGTM

While we are here do you have any thoughts on potentially adding some more info to the error in `check_running_system` saying that if this is not an Ubuntu-18.04 system you can bypass it by setting the appropriate environment variable at your own risk?

pzakha:
Good idea @sdimitro, I've updated the review with your suggestion.

sdimitro (approved):

sebroy (approved):

PR #197, opened 2021-10-20
TOOL-12355 linux-pkg: changes for Ubuntu 20.04

pzakha:
Note: this will only be integrated when transitioning master to Ubuntu 20.04.

The changes are fairly basic:
- Don't install gcc8 anymore because Ubuntu 20.04 is using gcc9, which has all the fixes we needed.
- Update the upstream git url where appropriate
- Change UBUNTU_DISTRIBUTION to "focal", which is used in a few checks.

## Testing
Those changes have been soaking on the focal branch for a few months

### DIFF
--- a/packages/linux-kernel-oracle/config.delphix.sh
+++ b/packages/linux-kernel-oracle/config.delphix.sh
@@ -18,7 +18,7 @@
 # shellcheck disable=SC2034
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/linux-kernel-oracle.git"
 
-UPSTREAM_GIT_URL="https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-oracle/+git/bionic"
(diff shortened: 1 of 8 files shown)


packages/linux-kernel-oracle/config.delphix.sh
```diff
@@ -18,7 +18,7 @@
 # shellcheck disable=SC2034
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/linux-kernel-oracle.git"
 
-UPSTREAM_GIT_URL="https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-oracle/+git/bionic"
+UPSTREAM_GIT_URL="https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-oracle/+git/focal"
```
sdimitro, line 21: Instead of hardcoding this, shall we do something like this:
```
UPSTREAM_GIT_URL="https://git.launchpad.net/~canonical-kernel/ubuntu/+source/linux-oracle/+git/${UBUNTU_DISTRIBUTION}"
```

PR #198, opened 2021-10-21
TOOL-12358 app-gate frontend build fails due to OOM

pzakha:
This adds a 4G swap file on the build VM when building packages.

This failure has started happening recently on the focal branch after syncing it up with master. It happens quite consistently but not every time so we are probably at the edge of memory usage.

While I think the front-end team should look a bit more into how to mitigate memory usage issues during the build, possibly via some custom build settings, having a swap file for mitigating memory usage spikes during the build might not be a bad idea in general. The drawback of adding the swap file is that it can mask some configuration issues where the build suddenly takes more memory.

See Jira for more info on the failure itself.

## Testing
- ab-pre-push on master (building virtualization): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6429/
- ab-pre-push on focal (building virtualization): http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6428/

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -26,7 +26,7 @@ logmust determine_default_git_branch
 # Update the sources.list file to point to our internal package mirror. If no
 # mirror url is passed in, then the latest mirror snapshot is used.
 #
-configure_apt_sources() {
+function configure_apt_sources() {
 	local package_mirror_url
 	local primary_url="$DELPHIX_PACKAGE_MIRROR_MAIN"
 	local secondary_url="$DELPHIX_PACKAGE_MIRROR_SECONDARY"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -81,6 +81,25 @@ configure_apt_sources() {
...
+	local size="4G"
+
+	if [[ ! -f "$swapfile" ]]; then
+		logmust sudo fallocate -l "$size" "$swapfile"
+		logmust sudo chmod 600 /swapfile
+		logmust sudo mkswap /swapfile
+	fi
+
+	if ! sudo swapon --show | grep -q "$swapfile"; then
+		logmust sudo swapon "$swapfile"
```
sebroy, line 99: I thought that mkswap should always be followed by swapon. I'm wondering what the extra conditional does.

PR #208, opened 2021-11-24
DLPX-78480 6.0/stage build needs python3.8 in order to release vSDK 4…

mrburke:
# Context:
The SDK team has been planning to release vSDK 4.0.0 (Python 3.8 support) with the 6.0.12.0 engine release. To do so, we need Python3.8 installed on the engine.

# Testing
Tested manually on a bootstrap 18.04 VM, and kicked off [an ab-pre-push job](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6638) - PASSED

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -23,6 +23,20 @@ logmust check_running_system
 logmust determine_default_git_branch
 
 #
+# In order to ship vSDK 4.0.0 with 6.0.12.0, we need python3.8 (and the
+# relevant distutils release). The following line gets the deadsnakes
+# (https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) repo from
+# which we install python3.8 in packages/virtualization/config.sh.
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):
I'm good with this landing as-is..

But I also want to mention (for posterity) that even though we plan to remove the external dependency being added here in a future version, we'll continue to depend on this external entity for all product builds of the version this goes into. So, for example, if this dependency ever ceased to exist, it'd cause hotfix build failures for all versions of the product that require this dependency.

This could be solved if `linux-pkg` wasn't versioned along with the product, for example.. and instead, if builds of all product versions used the `master` branch of `linux-pkg`, such that we could remove this dependency in the future, and that'd affect the build of prior product versions.. 

But that's not the build architecture we have today, so this change potentially does add risk to any future builds of product versions containing this change (i.e. hotfix builds).

sdimitro (approved):

PR #209, opened 2021-11-30
CP-5327 Simplify the fluentd build

brad-lewis:
Currently we build our own fluentd service in the delphix/omnibus-td-agent repo installing the gems needed for splunk.   We would like to eliminate this build and run the default td-agent and install the required gems at configuration time.   Here we eliminate the delphix td-agent buld and add the fluentd-gems package.   

This is a companion of [app-gate](https://jira.delphix.com/browse/CP-6676) and [appliance-build](CP-6676) changes.

Testing:
[ab-pre-push](http://selfservice.jenkins.delphix.com/job/devops-gate/job/master/job/appliance-build-orchestrator-pre-push/6632/)


sebroy (approved):

packages/fluentd-gems/config.sh
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bash
+# Copyright 2021 Delphix
+# Unless required by applicable law or agreed to in writing, software
```
prakashsurya, line 3: I think the copyright comment is mangled...?

package-lists/build/main.pkgs
```diff
@@ -14,6 +14,7 @@ delphix-rust
 delphix-sso-app
 drgn
 docker-python-image
+fluentd-gems
```
prakashsurya, line 17: you need to add the packages `config.sh` file, and add the package to `main.pkgs` in two independent PRs.. in order to avoid build failures.. the sequence need to be:

1. add the package's `config.sh` file
2. build the package in Jenkins
3. add the package to `main.pkgs`

If we don't perform (2) prior to (3), the Jenkins "combine-packages" job will fail to find the package for "fluentd-gems", and result in build failures (e.g. for `git-ab-pre-push`).. and we can't perform (2), until we've performed (1)..

So, while this is a bit of an awkward dance we need to perform, it's necessary given the current architecture, if we want to avoid impact to others (in the form of transient build failures)..

brad-lewis:
I removed the fluentd-gems/config.sh and opened a second PR for that.

prakashsurya (approved):

PR #215, opened 2021-12-02
Revert "DLPX-78480 6.0/stage build needs python3.8 in order to release vSDK 4…"

mrburke:
Reverting this change in favor of a different approach - see https://github.com/delphix/appliance-build/pull/636/

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -23,20 +23,6 @@ logmust check_running_system
 logmust determine_default_git_branch
 
 #
-# In order to ship vSDK 4.0.0 with 6.0.12.0, we need python3.8 (and the
-# relevant distutils release). The following line gets the deadsnakes
-# (https://launchpad.net/~deadsnakes/+archive/ubuntu/ppa) repo from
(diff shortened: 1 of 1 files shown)


prakashsurya (changes requested):
I don't think we want to revert this change; as I understand it's still needed.

Your appliance-build change here: https://github.com/delphix/appliance-build/pull/636/ .. will install python 3.8 on our development DE images, but it does not install it on our build server images. When linux-pkg runs to build the virtualization package, that is performed on a standard Ubuntu system, i.e. the `bootstrap-18-04` image on `dlpxdc.co`. Thus, in order to use python 3.8 during the package build, we still need to have some logic that installs it during the linux-pkg build (i.e. we need this logic).

Technically, I think we just need to add the new PPA here, as we'd then pull in python 3.8 via the addition of that package as a build dependency of the virtualization package; which you're adding here: http://reviews.delphix.com/r/75990

So, long story short, we still need to keep this change (at the very least, keep adding the new PPA).

mrburke:
I see, thank you! I'll close this revert

PR #216, opened 2021-12-09
TOOL-12604 linux-pkg's "push-merge.sh" script does not work for internal github repositories

pcd1193182:
This PR adds the ability to use the FETCH_GIT_TOKEN in this script. We pass it to the ls-remote call so that it can behave correctly on private github repositories. An accompanying change will be required to the devops repository to pass that variable to this script.

### DIFF
--- a/push-merge.sh
+++ b/push-merge.sh
@@ -76,8 +76,14 @@ set -o pipefail
 echo "Running: git rev-parse refs/heads/repo-HEAD-saved"
 saved_ref=$(git rev-parse refs/heads/repo-HEAD-saved) ||
 	die "Failed to read local ref refs/heads/repo-HEAD-saved"
+
+git_url="$DEFAULT_PACKAGE_GIT_URL"
+if [[ -n "$FETCH_GIT_TOKEN" ]] && [[ "$git_url" == https://github.com/* ]]; then
+	git_url="${git_url/https:\/\//https:\/\/${FETCH_GIT_TOKEN}@}"
+fi
+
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):

push-merge.sh
```diff
@@ -76,8 +76,14 @@ set -o pipefail
 echo "Running: git rev-parse refs/heads/repo-HEAD-saved"
 saved_ref=$(git rev-parse refs/heads/repo-HEAD-saved) ||
 	die "Failed to read local ref refs/heads/repo-HEAD-saved"
+
+git_url="$DEFAULT_PACKAGE_GIT_URL"
+if [[ -n "$FETCH_GIT_TOKEN" ]] && [[ "$git_url" == https://github.com/* ]]; then
```
prakashsurya, line 81: Actually, this should be `PUSH_GIT_TOKEN`, right? I'm not sure how they differ.. but I see `PUSH_GIT_TOKEN` used elsewhere.. e.g. in `push_to_remote()`

prakashsurya: or perhaps.. since this specific command is doing a read, `FETCH_GIT_TOKEN` is the correct one to use here?

pcd1193182: Since all we're doing in `ls-remote` is fetching a list of refs from the repository, I opted to use the `FETCH` token. The `PUSH` token is already handled in the actual push command, which happens further down in the script.

sebroy (approved):
The `docker/README.md` was updated by Jorge last week as he rolled out a new mechanism for deploying developer jenkins instances for testing. Would that help test this change?

push-merge.sh
prakashsurya: :+1: sounds good to me

PR #230, opened 2022-04-20
TOOL-13555 disable syncing upstream with zfs master branch

prakashsurya:


prakashsurya:
The code is a bit difficult to decipher, but I think with this code change landed to master, it'll disable this job from running: http://ops.jenkins.delphix.com/job/linux-pkg/job/master/job/update-package/job/zfs/

sebroy:
How does linux-pkg know what upstream repo and branch to sync 6.0/stage from?

prakashsurya:
> How does linux-pkg know what upstream repo and branch to sync 6.0/stage from?

This is only changing the `master` branch, the `6.0/branch` still has this logic.. so when we checkout the `6.0/branch` of linux-pkg via Jenkins, do the sync of upstream-master to downstream-6.0/stage, it'll have this logic..

Additionally, linux-pkg uses the `DEFAULT_GIT_BRANCH` variable to know which branch to sync upstream into.. on `master` that's configured to `master`, and on `6.0/stage` that's configured to `6.0/stage`.. So, I believe when we move the automation to sync upstream to `6.0/stage`, it'll have `UPSTREAM_GIT_BRANCH="master"` still set (since we won't backport this change), and `DEFAULT_GIT_BRANCH="6.0/stage"` set..

With all that said, IMO at least, this is all difficult to reason about.. and has strange inter-dependencies between different components and scripts in linux-pkg and the devops-gate.. so honestly, I can't say that I'm sure I understand it all correctly.. but based on my reading so far, this is how I think the Rube Goldberg machine works..

grwilson (approved):

sebroy (approved):

PR #245, opened 2022-08-30
DLPX-82803 [Backport of DLPX-81830] Upgrade Engine JDK to 8u345-b01

smeda-delphix:
Upgrade engine JDK to latest version available for 6.0.17.0. A newer version of JDK was released to fix a compiler bug after changes were done in master. So, updating the latest version to master as well.

ab-pre-push - http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3050/

### DIFF
--- a/packages/adoptopenjdk/config.sh
+++ b/packages/adoptopenjdk/config.sh
@@ -20,12 +20,12 @@ DEFAULT_PACKAGE_GIT_URL=none
 PACKAGE_DEPENDENCIES="make-jpkg"
 
 if [[ "$UPSTREAM_PRODUCT_BRANCH" == "master" ]]; then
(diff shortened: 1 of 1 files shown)


prakashsurya (approved):
Since you're making the two branches do the same thing, do we want to remove the conditional?

I'm OK leaving it, if we think we'll diverge again the next time we update the JDK.. and thus, we'd just add the conditional back.. but I wanted to ask..

smeda-delphix:
Yes, I would prefer leaving it for next time we update JDK.

rasantel (approved):

derek-woodham-delphix (approved):

PR #247, opened 2022-09-06
CP-6002 minimize Linux kernel build to only include necessary kernel modules

prakashsurya:


prakashsurya:
In my preliminary testing, this change on its own, shouldn't have any functional impact to the product.. but, it will allow us to make changes to the kernel repositories, such that we can tweak the resultant kernel config file that is used during the kernel build process.

I suggest we start with this change in isolation, such that it'll be easier to build, test, and verify any subsequent kernel changes.

prakashsurya:
`git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3121/) (w/ kernel build)

default-package-config.sh
```diff
@@ -155,6 +155,16 @@ function kernel_build() {
 	local build_deps_tool="apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends --yes"
 	logmust sudo mk-build-deps --install debian/control --tool "${build_deps_tool}"
 
+	#
+	# Here we update the configs used to control the kernel's build
+	# system. This is useful as it allows us to override various
+	# kernel config options via an OVERRIDES file, which we use to
+	# disable varous kernel modules that we don't need or want.
+	#
+	logmust debian/rules debian/control
```
sebroy, line 164: What does this do? This target appears to also be called by the `clean` target via dependency. Does it need to run before `updateconfigs`?

default-package-config.sh
```diff
@@ -155,6 +155,16 @@ function kernel_build() {
...
+	logmust fakeroot debian/rules updateconfigs "${debian_rules_args[@]}"
+	logmust fakeroot debian/rules clean "${debian_rules_args[@]}"
```
sebroy, line 166: Why do we now need to call `clean` whereas we didn't have to before?

sebroy (approved):

default-package-config.sh
prakashsurya: we may not need to.. I can remove it, and see how it goes..

default-package-config.sh
prakashsurya: I'll re-run the build, to grab the error message.. but when I hadn't run this step, the `updateconfigs` step would fail..

prakashsurya: `git-ab-pre-push` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3128/) .. this run is without this call.. lets see what happens..

prakashsurya: hm.. looks like that build is working.. when I run it manually on a clean VM, I see this:
```
$ fakeroot debian/rules updateconfigs
dh_testdir;
dh_testdir: error: "debian/control" not found. Are you sure you are in the correct directory?
make: *** [debian/rules.d/1-maintainer.mk:47: updateconfigs] Error 255
```

and I have to run:
```
$ ./debian/rules debian/control
```
to get past the issue..

but if the automation isn't hitting this issue, then I can remove this line..

sdimitro (approved):
^ For Seb's feedback above:
Some of these steps I copy pasted them in the script straight from the command line as I was doing different iterations outside of linux-pkg, so I think it'd make sense to check whether they are needed.

prakashsurya:
yea, seems like all we need is that one line to update the configs.. so that's what I'll go with.. I'll wait for the build to verify, but it's looking good so far.

PR #251, opened 2022-09-19
TOOL-15610 Add flag to buildpkg to cause the build to use locally-built dependencies

pcd1193182:
This PR adds a flag to buildpkg that allows the user to force the build process to use dependencies built locally in the repository instead of fetching them from s3. This is useful when doing rapid iteration or developing on new architectures, where the s3 versions do not work or do not contain necessary changes for testing.

Tested manually with and without the -l flag in use.

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -214,6 +214,7 @@ function check_git_ref() {
 #
 function stage() {
 	typeset hook=$1
+	shift 1
 
 	check_env PACKAGE
 	local stage_start=$SECONDS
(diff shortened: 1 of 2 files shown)


sebroy (approved):

lib/common.sh
```diff
@@ -712,6 +714,10 @@ function fetch_dependencies() {
 	for dep in $PACKAGE_DEPENDENCIES; do
 		echo "Fetching artifacts for dependency '$dep' ..."
 		get_package_prefix "$dep"
+		if [[ "$source" == "local" ]]; then
```
prakashsurya: since you're explicitly setting `source` to `s3` as the default.. should we tweak this to be something like:
```
if [[ "$source" == "local" ]]; then
    ...
elif [[ "$source" == "s3" ]]; then
    ....
else
    die "invalid source parameter specified: '$source'"
fi
```

I don't feel strongly, but the fact that we explicitly set the default as `s3`, and then never actually care if it's anything besides `local`, struck me as misleading... e.g. leads me to think that `s3` as a value behaves differently than `foo`, but I don't think that's true..

prakashsurya (approved):
looks reasonable to me..

prakashsurya (approved):

PR #253, opened 2022-09-22
TOOL-15610 Add flag to buildpkg to cause the build to use locally-built dependencies (#251) (fixed)

prakashsurya:

PR #252, opened 2022-09-22
TOOL-15610 Add flag to buildpkg to cause the build to use locally-built dependencies (#251) (revert)

prakashsurya:
This reverts #251.. I think the problem with that PR is the change from:
```
+       PACKAGE_DEPENDENCIES_METADATA="${PACKAGE_DEPENDENCIES_METADATA}$dep: $s3url\\n"
```
to
```
           PACKAGE_DEPENDENCIES_METADATA="" \
               "${PACKAGE_DEPENDENCIES_METADATA}$dep: $s3url\\n"
```

here's the failure message seen with #251
```
00:07:31.302  Fetched artifacts for 'host-jdks' from s3://snapshot-de-images/builds/jenkins-ops/linux-pkg/6.0/stage/build-package/host-jdks/post-push/8/
00:07:31.302  /var/tmp/jenkins/workspace/linux-pkg/6.0/stage/build-package/virtualization/post-push/linux-pkg/lib/common.sh: line 739: host-jdks: s3://snapshot-de-images/builds/jenkins-ops/linux-pkg/6.0/stage/build-package/host-jdks/post-push/8/\n: No such file or directory
00:07:31.302  Error: failed command 'fetch_dependencies s3'
```

IMO, since #251 isn't strictly necessary, we can revert this, and then just re-integrate with the fix in place.

PR #254, opened 2022-09-22
TOOL-15610 Add flag to buildpkg to cause the build to use locally-built dependencies (fix build breakage)

pcd1193182:
This is is a follow-on to the previous review that fixes a build breakage caused by this change. Because there is whitespace between the two halves of the string assignment, the second half is instead treated as a command to execute. This causes the build to fail. The change contained herein works to fix the issue, but I suspect it will cause the style checker to complain. Is there a more stylistic way to implement this?

### DIFF
--- a/lib/common.sh
+++ b/lib/common.sh
@@ -716,15 +716,13 @@ function fetch_dependencies() {
 		get_package_prefix "$dep"
 		case "$source" in
 		"local")
-			logmust cp -r "$WORKDIR/../../$dep/tmp/artifacts/" \
-			    "$dep/"
+			logmust cp -r "$WORKDIR/../../$dep/tmp/artifacts/ $dep/"
 			;;
 		"s3")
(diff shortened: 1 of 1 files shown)


prakashsurya:
I have two reviews out for this too.. #252 and #253 .. I'm OK moving forward with this PR, though.

prakashsurya (approved):

prakashsurya:
> but I suspect it will cause the style checker to complain

the `shfmt` PR check is green, so we're good as-is..

prakashsurya:
Also, in #253, I have a link that tested the virtualization build, which succeeded.. see [here](http://selfservice.jenkins.delphix.com/job/linux-pkg/job/6.0/job/stage/job/build-packages/job/pre-push/318/)

sumedhbala-delphix (approved):

PR #255, opened 2022-09-22
DLPX-83163 Add ARM support for adoptopenjdk package

pcd1193182:
Now that the OpenJDK build for arm64 has been added for artifactory, this PR enables ARM builds for the `adoptopenjdk` package.  Note that the removed code is identical to the untouched code, so I simplified the logic at the same time.

Tested manually on arm and amd64 bootstrap VMs.

### DIFF
--- a/packages/adoptopenjdk/config.sh
+++ b/packages/adoptopenjdk/config.sh
@@ -39,6 +39,12 @@ function prepare() {
 }
 
 function fetch() {
+	# We exit here rather than above in the architecture detection logic
+	# to deal with the fact that this file can, during test runs, be
+	# sourced on platforms where builds are not happening. list-packages
+	# sources the file to gather information about the package, and this
+	# is performed on jenkins and macos during test runs. Having the exit
+	# occur above causes those runs to fail.
 	if [[ -z "$_tarfile" ]]; then
 		echo "Invalid architecture detected" >&2
 		exit 1
--- a/packages/adoptopenjdk/config.sh
+++ b/packages/adoptopenjdk/config.sh
@@ -19,7 +19,7 @@
 DEFAULT_PACKAGE_GIT_URL=none
 PACKAGE_DEPENDENCIES="make-jpkg"
 
-case $(dpkg-architecture -q DEB_HOST_ARCH) in
+case $(dpkg-architecture -q DEB_HOST_ARCH 2>/dev/null || echo "none") in
 amd64)
 	_tarfile="OpenJDK8U-jdk_x64_linux_hotspot_8u345b01.tar.gz"
 	_tarfile_sha256="ed6c9db3719895584fb1fd69fc79c29240977675f26631911c5a1dbce07b7d58"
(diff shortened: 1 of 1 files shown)


packages/adoptopenjdk/config.sh
```diff
@@ -19,21 +19,30 @@
...
-	_tarfile_sha256="ed6c9db3719895584fb1fd69fc79c29240977675f26631911c5a1dbce07b7d58"
-	_jdk_path="/usr/lib/jvm/adoptopenjdk-java8-jdk-amd64"
-fi
+	;;
+arm64)
+	_tarfile="OpenJDK8U-jdk_aarch64_linux_hotspot_8u345b01.tar.gz"
+	_tarfile_sha256="c1965fb24dded7d7944e2da36cd902adf3b7b1d327aaa21ea507cff00a5a0090"
+	_jdk_path="/usr/lib/jvm/adoptopenjdk-java8-jdk-arm64"
+	;;
+*) ;;
```
sdimitro, line 33: shall we try to exit right away here instead of later down the line of execution?

sdimitro (approved):

packages/adoptopenjdk/config.sh
pcd1193182: That was the original design, but the way that list-packages works, it sources the file to get information about the package and dependencies. This causes the automated tests to fail, because they run on systems (e.g. macos) where the `dpkg-architecture` command isn't present. To work around this, I had to move the failure into the actual functions.

packages/adoptopenjdk/config.sh
```diff
@@ -19,21 +19,30 @@
 DEFAULT_PACKAGE_GIT_URL=none
 PACKAGE_DEPENDENCIES="make-jpkg"
 
-if [[ "$UPSTREAM_PRODUCT_BRANCH" == "master" ]]; then
```
prakashsurya, line 22: so, we left this conditional in, to make it easier to change master to be different the next time we need to do that..

but, I think.. given that we've effectively dropped support for the master branch, I'm not sure that idea still holds.. since we won't be changing the master version now that we don't support master..

so, I think this change is fine, to drop the conditional.. but I at least wanted to leave a bit of information in this PR as to why I think it's OK to drop it..

packages/adoptopenjdk/config.sh
prakashsurya: awkward, can you leave a brief comment in the script so we don't ask the same question later, if this ever gets modified.. copying what you have here into a comment should work..

packages/adoptopenjdk/config.sh
```diff
@@ -19,21 +19,30 @@
...
+case $(dpkg-architecture -q DEB_HOST_ARCH 2>/dev/null || echo "none") in
```
prakashsurya, line 22: out of curiosity, why do we need to "echo none"?

prakashsurya (approved):
LGTM

packages/adoptopenjdk/config.sh
pcd1193182: The testing scripts set `pipefail` when running `list-packages`; as a result, if this runs on a platform without dpkg-architecture, if the "or echo none" is omitted, the test run fails because the script bails out at this point.

prakashsurya: right.. that's what I figured.. I didn't see it in this script, but didn't look at any script that sources this one.. ok, makes sense, thanks.

PR #256, opened 2022-10-11
GHM-822 Build virtalization package with secrets

skinner-m-c:
Passes jumpbox information the virtualization build. 

Discussions about this work might want to take offline due to the nature of this work and this being a public repo.

### DIFF
--- a/packages/virtualization/config.sh
+++ b/packages/virtualization/config.sh
@@ -57,7 +57,7 @@ function build() {
 	#
 	logmust cd "$WORKDIR/repo/appliance"
 
-	local secretDbArgs=()
+	local args=()
 
 	# Because ant does not read the environment variables of the execution context, we set them as
 	# parameters only if they exist in the execution context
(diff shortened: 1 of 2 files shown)


packages/virtualization/config.sh
```diff
@@ -56,12 +56,61 @@ function build() {
...
+	# parameters only if they exist in the execution context
+
+	# Here we check for whether the environment variables are set and pass them along. We check for
+	# existence instead of emptiness to avoid adding a layer of interpretation.
+
+	# We use parameter expansion in the form of ${variable+nothing} which evaluates to the variable
+	# 'nothing' if 'variable' is not set. Because 'nothing' is not defined it evaluates to "" when 'variable'
+	# is not set. So [[ "" ]] is what is actually evaluated when 'variable' is not set.
+
+	if [[ ${SECRET_DB_USE_JUMPBOX+nothing} ]]; then
```
prakashsurya, line 72: Why not do something like:
```
if [[ -n "$SECRET_DB_USE_JUMPBOX" ]]; then
```
? 

this way, we wouldn't have to have the comment to explain the `+nothing` part... I see you mentioned this:

> We check for existence instead of emptiness to avoid adding a layer of interpretation.

but I don't think I understand what you mean...

packages/virtualization/config.sh
```diff
@@ -56,12 +56,61 @@ function build() {
...
+		secretDbArgs+=("-DSECRET_DB_USE_JUMPBOX=\"$SECRET_DB_USE_JUMPBOX\"")
+	fi
+
+	if [[ ${SECRET_DB_JUMP_BOX_HOST+nothing} ]]; then
```
prakashsurya, line 76: are all of these variables independent?

e.g. if one is set, should they all be set? or can each one be set independently?

I'm just wondering if we should have some error checking, based on which variables are set, and which ones are not.. like, if "this one" is set, throw an error if "these other ones" are not set..

I don't feel strongly, so feel free to drop this.. just thinking if there's a way to catch errors earlier, than moving forward with the build with incorrect parameters..

packages/virtualization/config.sh
```diff
@@ -56,12 +56,61 @@ function build() {
...
+			-Dbuild.legacy.resources.war=true \
+			"${secretDbArgs[@]}" \
+			all-secrets package
 	else
-		logmust ant -Ddockerize=true -DbuildJni=true all package
+		logmust ant \
+			-Ddockerize=true \
+			-DbuildJni=true \
+			"${secretDbArgs[@]}" \
+			all-secrets package
```
prakashsurya: maybe we should reduce this down into a single `ant` call.. doing something like:
```
args=()
if [[ -n "$DELPHIX_RELEASE_VERSION" ]]; then
    args+="-DhotfixGenDlpxVersion=$DELPHIX_RELEASE_VERSION"
    args+="-Dbuild.legacy.resources.war=true"
fi

logmust ant \
    -Ddockerize=true \
    -DbuildJni=true \
    "${args[@]}" \
    "${secretDbArgs[@]}" \
    all-secrets package
```

This is a bit unrelated to your change, since you didn't add the two `ant` calls.. but since you're adding some duplication to both sides of the conditional, maybe we should eliminate this duplication right now, too... what do you think?

maybe even use `args` for the `secretDbArgs` too.. I don't see any reason why we need to have two different variables..

prakashsurya (commented):
LGTM.. just some relatively minor comments/questions..

packages/virtualization/config.sh
skinner-m-c: With `test -n <variable>` makes no distinction between an undefined variable and a defined variable of length 0. It would not be forwarding the exact state it was given to the program it invokes. I should get the exact same behavior with these parameters as I would get if I were invoking the program directly in an interactive shell. If this script receives `SECRET_DB_USE_JUMPBOX=""` then the ant script should get `SECRET_DB_USE_JUMPBOX` set to `""`. It is also much easier to trace forwarded parameters when the layers they are pass through do nothing to them.

prakashsurya: is the case of `SECRET_DB_USE_JUMPBOX=""` something we actually care to pass through?

I see what you mean, but unless we care about that case, it just feel like unnecessary complexity...

prakashsurya (approved):

packages/virtualization/config.sh
skinner-m-c: Resolving from conversation conducted elsewhere. Re-post or contact me if I am mistaken.

packages/virtualization/config.sh
skinner-m-c: While I do think the general argument I have given is a good reason to make this distinction because there are some CLI programs I have used that only check to see if an environment variable is defined for changing behavior. In this instance `SECRET_DB_USE_JUMPBOX` only checks to see if the environment variables exists and not its contents. The standard behavior in many languages is have different behavior on whether an environment exists or not (e.g. Java and Python) and to throw an error or return null if it does not exist.

packages/virtualization/config.sh
skinner-m-c: Fixed

packages/virtualization/config.sh
```diff
@@ -56,14 +56,59 @@ function build() {
...
-			-Dbuild.legacy.resources.war=true all package
-	else
-		logmust ant -Ddockerize=true -DbuildJni=true all package
+		args+=("-DhotfixGenDlpxVersion=\"$DELPHIX_RELEASE_VERSION\"")
+		args+=("-Dbuild.legacy.resources.war=true")
 	fi
 
+	logmust ant \
+		"${args[@]}"
+	all-secrets package
```
prakashsurya: not a big deal, but can this be a single line now?

prakashsurya (approved):

packages/virtualization/config.sh
sebroy, line 72: Isn't this equivalent to `if [[ -n "$SECRET_DB_USE_JUMPBOX" ]]; then`, used elsewhere in this script?

sebroy (approved):

packages/virtualization/config.sh
prakashsurya: See [here](https://github.com/delphix/linux-pkg/pull/256#discussion_r993584275).. I asked basically the same question.

sonamkindy: I saw the earlier convo and I'm still not sure I follow why we'd need/want to use conditional parameter expansion. I opted to use the syntax `if [[ "$SECRET_DB_USE_JUMPBOX" ]]` in https://github.com/delphix/linux-pkg/pull/258.

skinner-m-c: https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash might help. We can discuss further outside of this code review if needed.

PR #258, opened 2022-10-14
GHM-824 Build masking and containerized masking packages with secrets

sonamkindy:
# Problem

masking and containerized masking package builds need to retrieve secrets from AWS. 

This is especially critical for when [DLPX-81645 Remove secrets from tip of dms-core-gate repo](https://delphix.atlassian.net/browse/DLPX-81645) changes are integrated.

# Testing

```sh
→ git-ab-pre-push -b "masking" --jenkins-url=http://selfservice.pkg-test.dcol2.delphix.com
Your build is at: http://selfservice.pkg-test.dcol2.delphix.com/job/appliance-build-orchestrator-pre-push/7/
```

Per http://selfservice.pkg-test.dcol2.delphix.com/job/linux-pkg/job/6.0/job/stage/job/build-package/job/masking/job/pre-push/5/console:
```
13:25:13  Success: Package masking has been built successfully.
```

# Future Improvements

See  [GHM-833](https://delphix.atlassian.net/browse/GHM-833).

### DIFF
--- a/packages/containerized-masking/config.sh
+++ b/packages/containerized-masking/config.sh
@@ -40,6 +40,32 @@ function build() {
 
 	logmust cd "$WORKDIR/repo"
 
+	if [[ "$SECRET_DB_AWS_ENDPOINT" ]]; then
+		export SECRET_DB_AWS_ENDPOINT="$SECRET_DB_AWS_ENDPOINT"
+	fi
+
+	# Using secrets proxy
+	if [[ "$SECRET_DB_USE_JUMPBOX" ]]; then
+		export SECRET_DB_USE_JUMPBOX="$SECRET_DB_USE_JUMPBOX"
+	fi
+	if [[ "$SECRET_DB_JUMP_BOX_HOST" ]]; then
+		export SECRET_DB_JUMP_BOX_HOST="$SECRET_DB_JUMP_BOX_HOST"
+	fi
+	if [[ "$SECRET_DB_JUMP_BOX_USER" ]]; then
+		export SECRET_DB_JUMP_BOX_USER="$SECRET_DB_JUMP_BOX_USER"
+	fi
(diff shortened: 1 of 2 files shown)


prakashsurya (approved):
Seems reasonable.. but, could we be consistent with #256.. it seems like both are doing the same thing, but with slightly different implementations.. even better, would probably be to move this logic into a function within the `common.sh` file, and have both call the common function.. 

I don't feel strongly, so I'm OK moving forward as-is (hence my approval), but I do think it'd be nice to consolidate the two and be consistent with each other (as much as would be reasonable, at least)..

Even in this one review, we have the duplication.. so having the common function might help us maintain this moving forward..

sonamkindy:
> Seems reasonable.. but, could we be consistent with #256.. it seems like both are doing the same thing, but with slightly different implementations.. even better, would probably be to move this logic into a function within the `common.sh` file, and have both call the common function..
> 
> I don't feel strongly, so I'm OK moving forward as-is (hence my approval), but I do think it'd be nice to consolidate the two and be consistent with each other (as much as would be reasonable, at least)..
> 
> Even in this one review, we have the duplication.. so having the common function might help us maintain this moving forward..

Sure thing, I can definitely consolidate that logic in a function in common.sh. The one discrepancy between the virtualization build config and masking/containerized masking build configs is that the former requires them as ant parameters whereas the latter requires them as environment variables. Are you amenable to me having a function just export the variables as environment variables? Since it's duplicated in both masking/containerized masking config shell scripts, it seems like it'd be the most useful to have them implement all of the necessary functionality.

**FYI**: raised [GHM-833](https://delphix.atlassian.net/browse/GHM-833) per slack convo with @prakashsurya.

skinner-m-c:
Might want to consider explicitly including the environment variables for the command. Exporting sends these variables to every command (e.g. the following rsync and cp commands and anything else added after this in the future).

rasantel (approved):

grodr (approved):

PR #260, opened 2022-10-20
GHM-822 Build virtalization package with secrets

skinner-m-c:
Corrections from previous PR include fixing the quoting around the arguments.

Also did more manual testing to confirm that linux-pkg invocation of the new targets with correct parameters passed in build the app-gate correctly.

See http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3606/ (already passed the linux-pkg phase).

### DIFF
--- a/packages/virtualization/config.sh
+++ b/packages/virtualization/config.sh
@@ -59,9 +59,6 @@ function build() {
 
 	local args=()
 
-	# Because ant does not read the environment variables of the execution context, we set them as
-	# parameters only if they exist in the execution context
-
 	# Here we check for whether the environment variables are set and pass them along. We check for
 	# existence instead of emptiness to avoid adding a layer of interpretation.
 
@@ -82,9 +79,9 @@ function build() {
 	fi
 
 	if [[ ${SECRET_DB_JUMP_BOX_PRIVATE_KEY+nothing} ]]; then
-	  if [[ ! -f "$SECRET_DB_JUMP_BOX_PRIVATE_KEY" ]]; then
-	    die "Jumpbox private key not found."
-	  fi
+		if [[ ! -f "$SECRET_DB_JUMP_BOX_PRIVATE_KEY" ]]; then
+			die "Jumpbox private key not found."
+		fi
 		args+=("-DSECRET_DB_JUMP_BOX_PRIVATE_KEY=$SECRET_DB_JUMP_BOX_PRIVATE_KEY")
 	fi
 
--- a/packages/virtualization/config.sh
+++ b/packages/virtualization/config.sh
@@ -56,13 +56,59 @@ function build() {
 	# Build the virtualization package
 	#
 	logmust cd "$WORKDIR/repo/appliance"
-	if [[ -n "$DELPHIX_RELEASE_VERSION" ]]; then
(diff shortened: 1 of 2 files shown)


packages/virtualization/config.sh
```diff
@@ -56,14 +56,62 @@ function build() {
...
-		logmust ant -Ddockerize=true -DbuildJni=true \
-			-DhotfixGenDlpxVersion="$DELPHIX_RELEASE_VERSION" \
-			-Dbuild.legacy.resources.war=true all package
-	else
-		logmust ant -Ddockerize=true -DbuildJni=true all package
+		args+=("-DhotfixGenDlpxVersion=$DELPHIX_RELEASE_VERSION")
+		args+=("-Dbuild.legacy.resources.war=true")
 	fi
 
+	echo "---- ${args[@]}"
```
prakashsurya: do we really want to be dumping sensitive information to stdout?

prakashsurya (approved):

grodr (approved):

packages/virtualization/config.sh
```diff
@@ -56,14 +56,60 @@ function build() {
 	# Build the virtualization package
 	#
 	logmust cd "$WORKDIR/repo/appliance"
+
+	local args=()
+
+	# Because ant does not read the environment variables of the execution context, we set them as
```
sebroy: I'm a bit skeptical of this comment on its face. Ant can read environment variables in general (e.g. https://coderanch.com/t/108737/build-tools/read-environment-variables-Ant). Is there something specific about our use of ant that prohibits this?

packages/virtualization/config.sh
```diff
@@ -56,14 +56,60 @@ function build() {
...
-		logmust ant -Ddockerize=true -DbuildJni=true \
-			-DhotfixGenDlpxVersion="$DELPHIX_RELEASE_VERSION" \
-			-Dbuild.legacy.resources.war=true all package
-	else
-		logmust ant -Ddockerize=true -DbuildJni=true all package
+		args+=("-DhotfixGenDlpxVersion=$DELPHIX_RELEASE_VERSION")
+		args+=("-Dbuild.legacy.resources.war=true")
 	fi
 
+	logmust ant "${args[@]}" all-secrets package
```
sebroy, line 108: Does this mean that we will log the private key and other args set above in the jenkins console log? I'm still not completely convinced that our ant build.xml couldn't read these from the environment using the `environment` property...

sebroy (changes requested):

packages/virtualization/config.sh
skinner-m-c: You can confirm this by checking the build at http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3606/ Jenkins redacts the secret.

packages/virtualization/config.sh
skinner-m-c: Comment removed. The value gained by redesigning to use environment variables is not worth delaying other work. Because this is an public discussion I am stopping here. If you wish to discuss further we can do it more openly in private.

sebroy (approved):

PR #261, opened 2022-11-07
TOOL-16515 Make dependency on "build-essential" explicit

prakashsurya:
We currently depend on the "build-essential" package being installed,
but don't explictly install it; we rely on it already being installed on
the system we're building on. This change makes this dependency
explicit, and we now pull it in if it's not already installed; as we do
for other packages we depend on.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -116,6 +116,7 @@ logmust sudo apt-get update
 # - jq is used to generate a JSON formatted metadata file by some packages.
 #
 logmust install_pkgs \
+	build-essential \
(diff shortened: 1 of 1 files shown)


prakashsurya:
- `git-ab-pre-push -b delphix-platform` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3720/)

sdimitro (approved):

sebroy (approved):

PR #263, opened 2022-11-11
QI-4423 linux-pkg changes for new branch names and release versioning

palash-gandhi:

### DIFF
--- a/branch.config
+++ b/branch.config
@@ -11,4 +11,4 @@
 # branching.
 #
 
-DEFAULT_GIT_BRANCH="6.0/stage"
+DEFAULT_GIT_BRANCH="develop"


prakashsurya:
I'm not even sure this setting makes sense.. e.g. if the release branch is ephemeral, when we branch from develop to release, this setting will still be set to "develop" even though it's on the "release" branch.. we might need to some how eliminate this setting..

prakashsurya:
I'm not 100% sure.. but I _think_ we workaround this for the current branch scheme.. by keeping the "6.0/release" branch around indefinitely, and then merging "6.0/stage" into "6.0/release" when we cutover for the new release.. this way, the merge commit doesn't change this file.. and we landed a commit to change the setting to "6.0/release" on the "6.0/release" branch..

but I could be wrong.. I haven't actually looked into how this works with the current branch scheme..

PR #264, opened 2022-11-14
TOOL-16737 Update "add_swap" function to work on Delphix buildserver

prakashsurya:


prakashsurya:
- `git-ab-pre-push -b bcc` is [here](http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/3779/)
- build using Delphix based buildserver is [here](http://ops.jenkins-psurya.dcol2.delphix.com/job/linux-pkg/job/6.0/job/stage/job/build-package/job/bcc/job/pre-push/20/)

setup.sh
```diff
@@ -86,18 +86,37 @@ function configure_apt_sources() {
...
-	if ! sudo swapon --show | grep -q "$swapfile"; then
-		logmust sudo swapon "$swapfile"
+	# Swap already enabled, nothing to do.
+	if sudo swapon --show | grep -q "$swapfile"; then
+		return
 	fi
+
+	if [[ "$rootfs" != "zfs" ]]; then
+		logmust sudo fallocate -l 4G "$swapfile"
+		logmust sudo chmod 600 "$swapfile"
```
sebroy, line 114: I'm not sure these commands make sense in the case where you're using a swap device. The device partition is a block device, so I'm not sure what fallocate will do in that case (fail, I'm guessing?). Also, you'd be changing the permissions on the `/dev` node, which is probably not what you want to do.

I think you can omit both of those commands, and the `mkswap` command will use the whole disk partition you point it at.

sebroy (changes requested):

setup.sh
prakashsurya: right.. but this block is only executed when the root filesystem is _not_ zfs.. i.e. when using the ubuntu buildserver..

so I agree.. I think you might have read the logic incorrectly though.. I'm specifically avoiding these calls when using the block device directly..

at least, that's the intention.. let me know if I got it wrong..

sebroy: You're right, I read this wrong. It's possible that I glanced at an earlier version of the PR that had the code organized slightly differently, but regardless, this looks right now.

sebroy (approved):

don-brady (approved):

PR #268, opened 2022-12-08
DLPX-83885 Clean up appliance build; resources_legacy.war can no longer be built

VenkatanadhanG:
### Problem
Jira Description: Clean up appliance build; resources_legacy.war can no longer be built
 
### Solution
Since legacy war can not be created, removing the args

### DIFF
--- a/packages/virtualization/config.sh
+++ b/packages/virtualization/config.sh
@@ -102,7 +102,6 @@ function build() {
 
 	if [[ -n "$DELPHIX_RELEASE_VERSION" ]]; then
 		args+=("-DhotfixGenDlpxVersion=$DELPHIX_RELEASE_VERSION")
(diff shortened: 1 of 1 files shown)


sebroy (approved):

prakashsurya (approved):
could you provide any pointers (if you have them) to explain why we no longer need `resources_legacy.war`? I'm unfamiliar with that artifact, so just curious...

manoj-joseph (approved):

VenkatanadhanG:
> could you provide any pointers (if you have them) to explain why we no longer need `resources_legacy.war`? I'm unfamiliar with that artifact, so just curious...

We have removed the backup code to support DSP on netty 3, this legacy war is generated if we choose a fallback to netty 3 for DSP
Since the code is deleted from app-gate, we are removing all its references in other projects.

PR #271, opened 2023-01-12
TOOL-17737 Update the virtualization package to point to GitHub

ghost:
# Problem

We are migrating the dlpx-app-gate to GitHub we need to point all of
our repositories to this version of the repository

# Solution

Make it so

# Bonus

A grep for `gitlab` revealed one more occurrence that should be
updated.


prakashsurya (approved):
I think we'll also have to update the `hotfix-update-remotes.py` script in the devops gate.

robinrye (approved):

grodr (approved):

PR #280, opened 2023-03-13
DLPX-85142 Unpin tar from linux-pkg on 6.0/stage

palash-gandhi:
This change unpins the `tar` package from 6.0/stage. It was added as part of DLPX-85006 in 6.0/release due to a CVE. We leave the infrastructure behind for future possibilities.
Not that branching for the new release has already completed, we are seeing some failures, unrelated to this change but that reminded me that we need to revert this on both these branches.

ab-pre-push -b misc-debs: http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/4834/

### DIFF
--- a/packages/misc-debs/config.sh
+++ b/packages/misc-debs/config.sh
@@ -47,9 +47,7 @@ SKIP_COPYRIGHTS_CHECK=true
 function fetch() {
 	logmust cd "$WORKDIR/artifacts"
 
-	local debs=(
-		"tar_1.30+dfsg-7ubuntu0.20.04.3_amd64.deb a3b1212570b10c192b3d49cffb68f909146d4b72a7104f213f4426c7e41d0e49"
-	)
+	local debs=()
 
 	local url="http://artifactory.delphix.com/artifactory/linux-pkg/misc-debs"
 


sonamkindy (approved):

prakashsurya:
We can do it this way, or we can preserve the `misc-debs` package, and just remove the `tar` package from it.. i.e. so it'd be empty.. the idea being, the "infrastructure" will be there if we need to do this in the future..

Do we have thoughts if it'd be useful to keep the package definition around, for potential future use? or remove it, and if needed later, it can be resurrected (like I did)?

prakashsurya (approved):
I'll approve as-is.. the notion if we want to keep the "misc-debs package infrastructure" around can be answered outside of this PR.. and we can always resurrect it later, even if we remove it in this PR..

palash-gandhi:
> We can do it this way, or we can preserve the `misc-debs` package, and just remove the `tar` package from it.. i.e. so it'd be empty.. the idea being, the "infrastructure" will be there if we need to do this in the future..
> 
> Do we have thoughts if it'd be useful to keep the package definition around, for potential future use? or remove it, and if needed later, it can be resurrected (like I did)?

I think it makes sense to leave the infra behind.

prakashsurya:
> I think it makes sense to leave the infra behind.

Let's verify that actually works, since I never tested it, with `git-ab-pre-push -b misc-debs`

PR #277, opened 2023-03-07
DLPX-85006 CVE-2022-48303 found in virtualization affected package tar_1.30+dfsg-7ubuntu0.20.04.2

prakashsurya:

PR #281, opened 2023-03-13
DLPX-85144 Revert "DLPX-85006 CVE-2022-48303 found in virtualization affected package tar_1.30+dfsg-7ubuntu0.20.04.2 (#277)" on 6.0/patch

palash-gandhi:
Similar to https://github.com/delphix/linux-pkg/pull/280

This change reverts https://delphix.atlassian.net/browse/DLPX-85006 which was a temporary way for us to pin the version of the tar package due to the CVE mentioned in the bug.
Not that branching for the new release has already completed, we are seeing some failures, unrelated to this change but that reminded me that we need to revert this on both these branches.


sonamkindy (approved):

prakashsurya:
I don't think we want to remove this on patch.. if we need to cut a patch release, we want that release to use the same tar package that we used on release.. if we revert this, we'll re-introduce the CVE if we need to make a patch release..

PR #287, opened 2023-05-17
TOOL-21304 Add support for delphix-go repository to linux-pkg

pcd1193182:
This change introduces the new delphix-go package into the linux-pkg repository so that the go compiler will be installed on dev VMs. It was tested in conjunction with corresponding changes to [appliance-build](url) and [devops-gate](https://github.com/delphix/devops-gate/pull/1124). ab-pre-push was run [here](http://selfservice.pd-jenkins.dcol2.delphix.com/job/appliance-build-orchestrator-pre-push/23/console) and the resulting VM had the GO sdk installed correctly and could be used for zfs builds.

### DIFF
--- a/packages/zfs/config.sh
+++ b/packages/zfs/config.sh
@@ -17,7 +17,7 @@
 # shellcheck disable=SC2034
 
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/zfs.git"
-PACKAGE_DEPENDENCIES="@linux-kernel delphix-rust dwarves"
+PACKAGE_DEPENDENCIES="@linux-kernel delphix-rust delphix-go dwarves"
 
 UPSTREAM_GIT_URL="https://github.com/openzfs/zfs.git"
 UPSTREAM_GIT_BRANCH="master"
@@ -56,6 +56,7 @@ function prepare() {
 		zlib1g-dev
 	logmust install_kernel_headers
 	logmust install_pkgs "$DEPDIR"/delphix-rust/*.deb
+	logmust install_pkgs "$DEPDIR"/delphix-go/*.deb
 	logmust install_pkgs "$DEPDIR"/dwarves/*.deb
 }
 
--- a/package-lists/build/main.pkgs
+++ b/package-lists/build/main.pkgs
@@ -9,6 +9,7 @@ challenge-response
 cloud-init
 crash-python
 crypt-blowfish
+delphix-go
 delphix-platform
 delphix-rust
 delphix-sso-app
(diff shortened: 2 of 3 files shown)


packages/zfs/config.sh
```diff
@@ -17,7 +17,7 @@
 # shellcheck disable=SC2034
 
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/zfs.git"
-PACKAGE_DEPENDENCIES="@linux-kernel delphix-rust dwarves"
+PACKAGE_DEPENDENCIES="@linux-kernel delphix-rust delphix-go dwarves"
```
prakashsurya, line 20: I think this is fine, but could potentially result in a package failure, between the point that we land this change, and the point that the initial package build of the `delphix-go` package completes..

We could split this.. i.e. add `delphix-go` in this PR.. then build the package.. then add the dependency in a follow on PR.. to avoid the potential for a build failure..

But I'm good as is.. if we see a build failure, it should be transient, so not a big deal.. just wanted to mention it.

prakashsurya (approved):

prakashsurya (commented):
I think you also need to add this package to the `package-lists/build/main.pkgs` file.. similar to the ZFS dependency comment.. that could cause build failures until the `delphix-go` package is successfully built.

Personally, I'm fine with that transient build breakage, due to how this is all designed.. but if we wanted to avoid it, we'd have to do the same follow-on PR approach, to add to `main.pkgs` after we add and build the package.

pcd1193182:
> I think you also need to add this package to the `package-lists/build/main.pkgs` file.. similar to the ZFS dependency comment.. that could cause build failures until the `delphix-go` package is successfully built.

I believe it is added in this PR as well. Personally I think I'm fine with transient failures as a result of everything spinning up, but it's good to raise that it's likely to happen so we can plan for it.  We can also probably mitigate it by integrating the delphix-go changes and devops-gate changes first?

prakashsurya:
> I believe it is added in this PR as well

shoot.. you're right :+1:

> Personally I think I'm fine with transient failures as a result of everything spinning up

cool with me..

> We can also probably mitigate it by integrating the delphix-go changes and devops-gate changes first?

yea, I'd try and land things in dependency order as much as you can..

1. devops-gate (no dependencies, repository already created)
2. delphix-go (independent, but PR checks required from 1)
3. linux-pkg (depends on build logic from 2)
4. appliance-build (depends on package from 3)

you might need to manually trigger the build after (3), so the package gets built the first time..

sebroy (approved):

PR #275, opened 2023-02-10
TOOL-19394 Remove branch.config from linux-pkg to allow pre-gitflow implementation

palash-gandhi:
<details open>
<summary><h2> Problem </h2></summary>
</details>
The plan for the branch rename project is to create the pre-gitflow branches 
preemptively to allow testing.
The linux-pkg repo has a file in each production branch of the repo
that contains the name of the branch. Most callers of scripts in this repo
set a `DEFAULT_GIT_BRANCH` environment variable that is consumed
by these scripts to determine the branch to operate on.

Here are some of these callers:
https://github.com/delphix/devops-gate/blob/main/jenkins/jobs/pipelines/linux_pkg_build.groovy#L29
https://github.com/delphix/devops-gate/blob/main/jenkins/jobs/pipelines/linux_pkg_build_package.groovy#L45
https://github.com/delphix/devops-gate/blob/main/jenkins/jobs/pipelines/linux_pkg_check_updates.groovy#L26

The plan for the branch rename project is to create mirrored branches i.e. `develop` to be a mirror of `6.0/stage`
and so on. To avoid having branch-specific information in the branch, this change proposes to remove this file since
most of the consumers of this repo set this env var.

<details open>
<summary><h2> Solution </h2></summary>
Remove the file and all consumers of it.

git-utils change: https://github.com/delphix/git-utils/pull/77
</details>

<details open>
<summary><h2> Testing Done </h2></summary>

generate-jobs: http://ops.jenkins-palashgandhi.dcol2.delphix.com/job/generate-jobs/23/

build-packages: http://ops.jenkins-palashgandhi.dcol2.delphix.com/job/linux-pkg/job/6.0/job/stage/job/build-packages/job/pre-push/2/
</details>

PR #291, opened 2023-06-06
TOOL-21507 linux-pkg should fetch releases from new release dataset paths

palash-gandhi:
<details open>

The pregitflow project changed the dataset paths of shipped releases from 
http://linux-package-mirror.delphix.com/release/ to http://linux-package-mirror.delphix.com/releases/.

This caused a regression because the linux-pkg repo used the `DEFAULT_GIT_BRANCH` as the path of the dataset to fetch from because they were both the same `release/A.B.C.D` but now they have changed. The "branches" (tags, really) are still following the old scheme `release/A.B.C.D` but the datasets are not under `releases/A.B.C.D`.

So, I modified the jenkins job in https://github.com/delphix/devops-gate/pull/802 to pass the new scheme to linux-pkg,
and it worked for most build-pkg jobs of shipped releases until now when I tried to ship a hotfix of the azure kernel package: http://selfservice.jenkins.delphix.com/job/linux-pkg/job/custom/job/build-package/job/custom/221/console
```
00:04:06.912  Running: [token passed] git fetch https://github.com/delphix/zfs.git --no-tags +releases/7.0.0.0:repo-HEAD --depth=1
00:04:07.251  fatal: couldn't find remote ref releases/7.0.0.0
```
The reason why this has likely only come to light right now is because the `virtualization` package does not have dependencies that needed to be rebuilt.
```
00:00:23.132  Final build list after applying BUILD_POLICY and REBUILD_DEPENDENCIES policies:
00:00:23.132    [virtualization]
```
vs in my case:
```
00:00:44.823  Final list of packages to build:
00:00:44.823    [linux-kernel-azure, connstat, delphix-kernel, zfs]
```
</details>

<details open>

The fix should be to revert the jenkins job to continue using the true branch scheme `release/A.B.C.D` instead of `releases/A.B.C.D` and modify linux-pkg to not conflate the value to fetch packages from the mirror.
Devops-gate review: https://github.com/delphix/devops-gate/pull/1271
</details>

<details open>

http://selfservice.jenkins-palashgandhi.dcol2.delphix.com/job/hotfix-build/14/console
Build a virtualization HF: http://selfservice.jenkins-palashgandhi.dcol2.delphix.com/job/hotfix-build/16/console
</details>

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -33,7 +33,7 @@ function configure_apt_sources() {
 	if [[ -z "$primary_url" ]] || [[ -z "$secondary_url" ]]; then
 		local latest_url="http://linux-package-mirror.delphix.com/"
 		if is_release_branch; then
-			package_mirror_url="${latest_url}${DEFAULT_GIT_BRANCH}"
+			package_mirror_url="${latest_url}releases/${DELPHIX_RELEASE_VERSION}"
 		else
 			latest_url+="${DEFAULT_GIT_BRANCH}/latest/"
 			package_mirror_url=$(curl -LfSs -o /dev/null -w '%{url_effective}' \


setup.sh
```diff
@@ -33,7 +33,7 @@ function configure_apt_sources() {
 	if [[ -z "$primary_url" ]] || [[ -z "$secondary_url" ]]; then
 		local latest_url="http://linux-package-mirror.delphix.com/"
 		if is_release_branch; then
-			package_mirror_url="${latest_url}${DEFAULT_GIT_BRANCH}"
+			package_mirror_url="${latest_url}releases/${DELPHIX_RELEASE_VERSION}"
```
prakashsurya, line 36: not necessarily specific to this change, but are we trying to move away from the "upstream product branch" concept, and instead specify the "delphix version"? e.g. even in the case of building an unreleased version?

has the package mirror URLs moved away from using branch names, to using delphix version numbers?

prakashsurya: oh, wait.. this is specific to when `is_release_branch` is set.. so then, my question doesn't really apply anymore.. I'll resolve this..

grodr (approved):

setup.sh
prakashsurya, line 36: What happens when you run a linux-pkg build manually? E.g. did you test something like:

1. Clone a buildserver VM
2. Checkout linux-pkg on the VM
3. Run `./setup.py`
4. Run `./buildpkg.sh`

prakashsurya: I guess for this manual case, most folks won't be trying to build a release tag.. so maybe this is fine.. not even sure if that manual workflow still works today..

prakashsurya (approved):
I'm OK with this as-is.

prakashsurya (approved):
I'm OK with this as-is.

setup.sh
palash-gandhi: yeah you'd have to have the correct environment vars like DEFAULT_GIT_BRANCH, PACKAGE_GIT_BRANCH and more.

prakashsurya: IIRC, all those get set automatically.. or at least they used to.. at least for (the old) `master` based builds..

palash-gandhi: Not anymore since https://github.com/delphix/linux-pkg/pull/275. But yeah, what I meant was that the repo requires a set of environment vars to be set and `DELPHIX_RELEASE_VERSION` is already set in our automation afaict.

PR #293, opened 2023-06-09
DLPX-85175 upgrade kernel to better leverage bpftrace and BTF

prakashsurya:


palash-gandhi (approved):
For my knowledge, Is UBUNTU_DISTRIBUTION set to something else when building the new kernel?

prakashsurya:
> For my knowledge, Is UBUNTU_DISTRIBUTION set to something else when building the new kernel?

No, don't think so.. I haven't changed that at all.. I believe that won't change until we move to the new LTS..

sdimitro (approved):

PR #296, opened 2023-07-03
RENG-145 Update config.sh fetch and pass upstream branch to the linux-pkg ant task

shishaktdlpx:
<details open>

This PR is a part of the PR (https://github.com/delphix/devops-gate/pull/1406) and https://github.com/delphix/dlpx-app-gate/pull/802.
This change is required to pickup the build.branch automatically.
</details>

<details open>

Config.sh has been modified where the ant command executes.
</details>

<details>

`07:33:53  Running: ant -Dbuild.branch=develop -DSECRET_DB_USE_JUMPBOX=true -DSECRET_DB_JUMP_BOX_HOST=cred-proxy.ops.delphix.com -DSECRET_DB_JUMP_BOX_USER=ec2-user -DSECRET_DB_JUMP_BOX_PRIVATE_KEY=**** -DSECRET_DB_AWS_ENDPOINT=vpce-064c6c9e8160b1cd9-b3i8sy8a.secretsmanager.us-west-2.vpce.amazonaws.com -Ddockerize=true -DbuildJni=true -DhotfixGenDlpxVersion=14.0.0.0 all-secrets package
`
(http://selfservice.jenkins-shishaktkumar.dcol2.delphix.com/job/linux-pkg/job/develop/job/build-package/job/virtualization/job/pre-push/3/consoleFull)
</details>

### DIFF
--- a/packages/virtualization/config.sh
+++ b/packages/virtualization/config.sh
@@ -66,6 +66,8 @@ function build() {
 	# 'nothing' if 'variable' is not set. Because 'nothing' is not defined it evaluates to "" when 'variable'
 	# is not set. So [[ "" ]] is what is actually evaluated when 'variable' is not set.
 
+	args+=("-Dbuild.branch=$DEFAULT_GIT_BRANCH")
+
 	if [[ ${SECRET_DB_USE_JUMPBOX+nothing} ]]; then
 		args+=("-DSECRET_DB_USE_JUMPBOX=$SECRET_DB_USE_JUMPBOX")
 	fi


packages/virtualization/config.sh
```diff
@@ -66,6 +66,10 @@ function build() {
 	# 'nothing' if 'variable' is not set. Because 'nothing' is not defined it evaluates to "" when 'variable'
 	# is not set. So [[ "" ]] is what is actually evaluated when 'variable' is not set.
 
+	if [[ ${UPSTREAM_PRODUCT_BRANCH+nothing} ]]; then
```
km4dlpx: @shishaktdlpx How does this variable get the upstream branch value?

shishaktdlpx: @km4dlpx  In the `linux_pkg_build_package.groovy` the **$UPSTREAM_PRODUCT_BRANCH** is already defined in their `env` so it will take it from there. 
An example of that can be found here. http://selfservice.jenkins-shishaktkumar.dcol2.delphix.com/job/linux-pkg/job/develop/job/build-package/job/virtualization/job/pre-push/1/console 
`17:16:43  Running: ant -Dbuild.branch=develop -DSECRET_DB_USE_JUMPBOX=true`

jorge-morse (approved):

packages/virtualization/config.sh
palash-gandhi: Shouldn't this be `DEFAULT_GIT_BRANCH`?

packages/virtualization/config.sh
```diff
@@ -66,6 +66,10 @@ function build() {
...
+		args+=("-Dbuild.branch=$UPSTREAM_PRODUCT_BRANCH")
+	fi
```
palash-gandhi: Shouldn't this raise an exception if the branch is not set?

palash-gandhi (changes requested):

packages/virtualization/config.sh
shishaktdlpx: The code to raise exception is already there in the pipeline.

packages/virtualization/config.sh
shishaktdlpx: Updated.

packages/virtualization/config.sh
palash-gandhi: This code can also be run outside of the pipeline and should have it's own `check_env`. See other instances of `check_env` in this repo.

shishaktdlpx: https://github.com/delphix/linux-pkg/blob/203e5932d03f53e3ecadbaa5fb8b58d96c545b34/buildpkg.sh#L70

the code for check_env is already in `buildpkg.sh` before triggering the build function. I don't think we should write extra `check_env` inside `config.sh` .

palash-gandhi: @shishaktdlpx So if this variable is always set, we can remove the `if` statement, right?

shishaktdlpx: Yes I think so. But there is no harm to let it be.

palash-gandhi: If there is a caller of this function that does not set the env var, the function will not fail early. I think we should remove it.

shishaktdlpx: Done.

palash-gandhi (approved):
Please retest your final diff

PR #297, opened 2023-07-25
DLPX-87116 Add cargo license tool to zfs build

pcd1193182:
### Problem
In order to facilitate keeping our BOM up to date, we need to either update the object agent THIRDPARTYLICENSES file in the repo, or regenerate it at build time. The latter is significantly less development upkeep. This requires that our build environment have the license bundling tool.

### Solution
We add the package to the setup process for building the zfs package.

### Testing Done
ab-pre-push with this and [the related ZFS changes](https://github.com/delphix/zfs/pull/1066): http://selfservice.jenkins.delphix.com/job/appliance-build-orchestrator-pre-push/6269/console

### DIFF
--- a/packages/zfs/config.sh
+++ b/packages/zfs/config.sh
@@ -56,6 +56,7 @@ function prepare() {
 		zlib1g-dev
 	logmust install_kernel_headers
 	logmust install_pkgs "$DEPDIR"/delphix-rust/*.deb
+	logmust cargo install cargo-bundle-licenses
 	logmust install_pkgs "$DEPDIR"/delphix-go/*.deb
 	logmust install_pkgs "$DEPDIR"/dwarves/*.deb
 }


packages/zfs/config.sh
```diff
@@ -56,6 +56,7 @@ function prepare() {
 		zlib1g-dev
 	logmust install_kernel_headers
 	logmust install_pkgs "$DEPDIR"/delphix-rust/*.deb
+	logmust cargo install cargo-bundle-licenses
```
prakashsurya, line 59: I'm not sure this PR should be blocked on this concern.. but I wonder what this implies in terms of build reproducibility.. and curios if we should be pulling this from some internal resource (e.g. a mirror on artifactory)..

For example, if we build ZFS for a hotfix.. the version of `cargo-bundle-licenses` we pull down for the build may not match what was originally used to build the old release.. this is in contrast to how we have the APT mirror pinned, so that rebuilding for a hotfix should pull in all the same (old) versions of the build dependency packages.. and thus, rebuild the package in the same way that we originally did for the older release..

Additionally.. we don't have to depend on an external resource to do the build (e.g. Ubuntu's APT repository), since we only use our internal mirror..

I'm good with this PR as-is.. just some things that come to mind as I look at this.. and it's probably a question more broadly about how we manage cargo dependencies in general, and not necessarily specific to this PR..

prakashsurya (approved):
LGTM

packages/zfs/config.sh
pcd1193182: Agreed on all counts.

sdimitro (approved):

PR #318, opened 2024-07-31
TOOL-25848 add DCT package to linux-pkg

prakashsurya:
<details open>
To support installing the "delphix-dct" package on Delphix engines, we need to have the DCT package available to be installed by the build system.
</details>

<details open>
While the actual building of the DCT package will not be handled by linux-pkg, we're choosing to use linux-pkg to make the pre-built DCT package available to the build system. Thus, the config.sh file for DCT is simply downloading the package from S3, rather than build the package from scratch.
</details>

<details>

Used "git-linux-pkg build" to verify the DCT package config.sh works.
```
PACKAGE dct: STAGE fetch STARTED
Running: fetch
Running: aws s3 cp s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/latest .
download: s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/latest to ./latest
Running: rm -f latest
Running: cd artifacts
Running: aws s3 sync s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/1970 .
download: s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/1970/SHA256SUMS to ./SHA256SUMS
download: s3://snapshot-de-images/builds/jenkins-ops/dct/develop/post-push/1970/delphix-dct_20.0.0-1722250787357.7788e36119ec_amd64.deb to ./delphix-dct_20.0.0-1722250787357.7788e36119ec_amd64.deb
Running: sha256sum -c SHA256SUMS
delphix-dct_20.0.0-1722250787357.7788e36119ec_amd64.deb: OK
PACKAGE dct: STAGE fetch COMPLETED in 40 seconds
```
</details>

### DIFF
--- a/packages/dct/config.sh
+++ b/packages/dct/config.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+#
+# Copyright 2024 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL=none
+SKIP_COPYRIGHTS_CHECK=true
+
+#
+# DCT builds use this URL for storing their artifacts, so we have it hardcoded here.
+#
+DCT_S3_DIR="s3://snapshot-de-images"
+DCT_LATEST_PREFIX="builds/jenkins-ops/dct/develop/post-push/latest"
+
+function fetch() {
+	logmust aws s3 cp "$DCT_S3_DIR/$DCT_LATEST_PREFIX" .
+
+	DCT_PACKAGE_PREFIX=$(cat latest)
+	logmust rm -f latest
+
+	logmust cd artifacts
+	logmust aws s3 sync "$DCT_S3_DIR/$DCT_PACKAGE_PREFIX" .
+	logmust sha256sum -c SHA256SUMS
+}
+
+function build() {
+	return
+	# Nothing to do, all the logic is done in fetch().
+}
--- a/package-lists/build/main.pkgs
+++ b/package-lists/build/main.pkgs
@@ -9,6 +9,7 @@ challenge-response
 cloud-init
 crash-python
 crypt-blowfish
+dct
 delphix-go
 delphix-platform
 delphix-rust


packages/dct/config.sh
```diff
@@ -0,0 +1,42 @@
...
+# shellcheck disable=SC2034
+
+DEFAULT_PACKAGE_GIT_URL=none
+SKIP_COPYRIGHTS_CHECK=true
+
+#
+# DCT builds use this URL for storing their artifacts, so we have it hardcoded here.
+#
+DCT_S3_DIR="s3://snapshot-de-images"
+DCT_LATEST_PREFIX="builds/jenkins-ops/dct/develop/post-push/latest"
```
palash-gandhi, line 26: Should we be pinning this here? How would this work for hotfixes or patch releases? For my knowledge, can you please clarify the following cases?
- Building a hotfix unrelated to a DCT fix (unused, correct?)
- Building a patch unrelated to a DCT fix (unused, correct?)
- Building a hotfix related to a DCT fix assuming we plan to leverage existing hotfix automation. Wouldn't this mean that it will always use the "latest" DCT from the main branch?
- Building a patch related to a dCT fix assuming we plan to leverage existing patch automation. Ditto.
Also, this URL needs to change based on the branch being built right? I think we should also revisit the idea of not using the linux-pkg framework for DCT appliances.

prakashsurya: > Building a hotfix unrelated to a DCT fix (unused, correct?)

IIRC, it'll use w/e DCT package was used/pinned when that release was created.

> Building a patch unrelated to a DCT fix (unused, correct?)

I think for this case, it will pull the "latest" DCT package, since we do a complete rebuild of the appliance.

This is (arguably) by design..  since DCT doesn't follow the same develop/release/patch branching, we can't "freeze" it like we do other repositories, via branching..

If it's not acceptable to always pull the latest DCT package, even for patch releases, then this wont be sufficient.

I'd argue it's fine, because: 1. I think it might actually be reasonable to pull the latest, even for a patch release.. and 2. we don't plan to release the DCT appliance the same way as a regular appliance, so the "latest" package won't even be installed for normal patch releases (as far as I understand so far)..

If we do need to "freeze" DCT for patches, etc.. then we need to do something else.. but I'm not sure we need to support that, so far.

> Building a hotfix related to a DCT fix assuming we plan to leverage existing hotfix automation

Based on my understanding, we won't support hotfixes for the DCT appliance.. at least, not the same way we do them today for non-DCT appliances..

If we did need to support them, then I think we'd need to do something a little different..

> Wouldn't this mean that it will always use the "latest" DCT from the main branch?

Correct, that's by design.

> Also, this URL needs to change based on the branch being built right?

No, I specifically have it always pull the latest.. DCT doesn't have the concept of different branches..

> Building a patch related to a dCT fix assuming we plan to leverage existing patch automation.

We won't use the existing patch automation.. the DCT release process will have it's own way to build "release" and "patch" DCT packages, and have it's own automation to coordinate those packages with the build process.. 

> I think we should also revisit the idea of not tying the linux-pkg framework for DCT appliances.

Sure, what do you propose? And would that proposal support the various cases you were asking about? e.g. hotfixes, patches, etc..

I think it could be valid to have appliance-build always pull the lastest DCT package, like we've mentioned.. that likely won't work the same for the edge cases, but I don't think those apply for DCT..

With all that said, I think another approach to do this, might be to pull the DCT package into the package mirror that we generate.. e.g. treat this more like a "3rd party package", like we do for postgres and some others.. where the package mirror pulls it from "somewhere", and makes it available to the build.. in this case, it'd pull it from S3..

I think part of the issue is much of this logic assumes we're building or rebuild the package, and branch it along with the product.. but that's not the approach we're taking with DCT.. where DCT is more like a "3rd party package", with it's own branching and release process.. and we just consume their latest package artifact.. we don't do hotfixes/patches/etc. of it (at least, aren't currently planning to).. I think it'd still be possible to do hotfixes/etc. much like it's possible to do with our "misc-debs" package, but doesn't come with all the same branching/building/etc. infrastructure..

palash-gandhi: > No, I specifically have it always pull the latest.. DCT doesn't have the concept of different branches..

What I meant was different "appliance" branches will require different DCT versions to be installed. For e.g, a build of the "develop" branch of the appliance may require bits from the "main" of the DCT repo. a build of the "release" branch of the appliance may require a different version of the DCT repo. I remember this discussion about these combinations

palash-gandhi: We are saying that we do need to support existing hotfix automation or patch automation for DCT appliances. But we do plan on using the linux-pkg framework, appliance-build and appliance-snapshots. Our non-DCT appliances are generally built in this flow linux-pkg -> appliance-build -> appliance-snapshots. 

You're proposing that we "inject" DCT into the linux-pkg step whereas I am proposing that we "inject" DCT at the appliance-build step. That way, DCT does not need to be included in the linux-pkg framework and a full different set of flavors of our build & release automation (dct-appliance-build, dct-appliance-snapshots, dct-release-orchestrator, dct-hotfix-build etc) can be created. It's hard to visualize and validate the current set of changes and how these future problems will be solved. If we want to integrate this as-is for now, I am okay with that too. The only problem is that the linux-pkg repo branches with each appliance release and introducing this change may introduce some debt if we choose to take another route in a few months.

prakashsurya: > For e.g, a build of the "develop" branch of the appliance may require bits from the "main" of the DCT repo. a build of the "release" branch of the appliance may require a different version of the DCT repo.

correct.. the plan is to accomplish this via the new (doesn't exist yet) DCT release jenkins job, that will pass in the S3 URL to the DCT package they wish to use for that specific release.. this is already supported by our linux-pkg jenkins jobs, IIRC..

so this logic always pulls the latest.. and any other job that doesn't want the latest, will need to pass in the S3 URL explicitly..

and it'll be up to the DCT team to build different packages for their different purposes.. and store them in unique S3 URLs.. and also pass those URLs to their release automation, that is yet to be written..

prakashsurya: > You're proposing that we "inject" DCT into the linux-pkg step whereas I am proposing that we "inject" DCT at the appliance-build step. 

like this: https://github.com/delphix/appliance-build/pull/765/commits/e69769bd53743bba38cbf1286f10f931069c630a ?

prakashsurya: > Our non-DCT appliances are generally built in this flow linux-pkg -> appliance-build -> appliance-snapshots.

correct, and this would remain for DCT appliances.. there's really nothing special about a "DCT" appliance.. it's just we need the DCT package to be available to the build system, so the DCT package can be installed..

I don't think we're supporting the release/patch/hotfix process and builds for DCT.. in a sense, it's a 3rd party package.. and just like packages we pull from Ubuntu, or postgres, or wherever.. we don't 'support" branching/hotfixes for those either.. (we can do it, just might not be a oneliner)..

All I'm trying to accomplish, is a way for the latest DCT package to be available to be installed via appliance-build.. and also a way for the DCT to specific which S3 URL to use (i.e. not the latest) for their release.. which I think this change accomplishes..

And also, not affecting our non-DCT builds.. which I think this also accomplishes.. 

> and a full different set of flavors of our build & release automation (dct-appliance-build, dct-appliance-snapshots, dct-release-orchestrator, dct-hotfix-build etc) can be created.

I don't think there's any plans to do this, and don't think we want to separate it out like that.. the DCT variant is no different than any other variant we have today, so I don't think we should do anything special for it.. it should work just like all our other variants..

the only caveat, is we need to make this "3rd party" DCT package available to the build somehow.. but that's tangential to running the build, or making a release, or dcenter image, etc..

palash-gandhi: I was proposing that the future, DCT flavor of the appliance-build jenkins job should include an extra parameter for the DCT package. Then, we add logic in the `appliance-build` repo (in the ansible role specific to the DCT package, or elsewhere) to download this package.

palash-gandhi: > the DCT variant is no different than any other variant we have today, so I don't think we should do anything special for it.. it should work just like all our other variants..

It is, in the sense that it is built & released independently of the regular appliance. As such, we have to create new flavors of the jenkins jobs.

prakashsurya: > I was proposing that the future, DCT flavor of the appliance-build jenkins job should include an extra parameter for the DCT package. Then, we add logic in the appliance-build repo (in the ansible role specific to the DCT package, or elsewhere) to download this package.

Have you checked the commit I linked to? this one: https://github.com/delphix/appliance-build/commit/e69769bd53743bba38cbf1286f10f931069c630a

I think I'd be OK going with this approach, which might be more inline with what you're thinking? i.e. have appliance-build download the latest DCT package by default.. but allow that to be overridden, to allow a different S3 location to be used.. e.g. for the DCT release build, etc..

palash-gandhi: Would this change still be required then?

prakashsurya: If we go with that approach, I'll close this.

Honestly, I think I'm leaning towards that approach, too.. I'll go ahead and close this.. let me know if you disagree..

prakashsurya:
Closing in favor of having appliance-build pull in the package itself...

PR #325, opened 2024-10-03
DLPX-92534 linux-pkg changes for LTS upgrade to 24.04

palash-gandhi:
<details open>

This PR contains changes required for the Ubuntu LTS upgrade to 24.04 and as such targets the `os-upgrade` branch. It mainly removes a few internal sources in favor of upstream packages and also enables building the remaining packages on 24.04.
</details>

<details open>

The following are the changes, along with explanations for each:
1. `bpftrace`, `bcc`, `drgn`, `fio`, `gdb-python`, `libkdumpfile` and `mold`: All of these packages also can be sourced directly from their respective upstream sources (dependent changes WIP)
2. New kernel build parameters like `do_dbgsym_package=True`, `do_tools_common=false` and `do_tools_host=false` added with explanations.
3. `llvm-12` is not available on 24.04. As such update `connstat` and `zfs`'s build dependency to depend on `llvm-14` instead.
4. `adoptopenjdk`: Prior to this PR, the JDK was sourced and downloaded from our internal artifactory. There is no reason to not install this via apt.
5. `zfs` also requires additional packages:
    - `dh-dkms`
    - `libpam0g`: The bootstrap image that the current + future OS upgrade projects will initially use is based on a vanilla Ubuntu image, and not a Delphix-based buildserver. Hence, the PAM headers need to be installed as a build dependency of ZFS
    - `libtirpc-dev`
6. Add mirror selection logic to select v2 mirrors for versions > 2025.2.

</details>

<details open>

https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/os-upgrade/job/build-packages/job/pre-push/51/
</details>


jwk404 (approved):

prakashsurya:
> Prior to this PR, the JDK was sourced and downloaded from our internal artifactory. There is no reason why we cannot install this via apt.

can you clarify why it's OK to remove the adoptopenjdk package? why did we need it to begin with?

palash-gandhi:
> > Prior to this PR, the JDK was sourced and downloaded from our internal artifactory. There is no reason why we cannot install this via apt.
> 
> can you clarify why it's OK to remove the adoptopenjdk package? why did we need it to begin with?

From https://perforce.slack.com/archives/C07HB5Y7K8S/p1729097131001349, it sounds like it was an old process that we stuck to and there's no reason to maintain a copy of the package in artifactory.

prakashsurya:
Thanks, OK.. I'd defer to Seb, so if he's OK with it, then seems reasonable to me.. perhaps when we started down that path, there wasn't a JDK 8 available from ubuntu..

sebroy (approved):

packages/containerized-masking/config.sh
```diff
@@ -29,17 +29,15 @@ source "$PWD/lib/common.sh"
...
 function prepare() {
-	logmust install_pkgs "$DEPDIR"/adoptopenjdk/*.deb
+	echo "Nothing to prepare"
 }
 
 function build() {
 	export JAVA_HOME
-	JAVA_HOME=$(cat "$DEPDIR/adoptopenjdk/JDK_PATH") ||
-		die "Failed to read $DEPDIR/adoptopenjdk/JDK_PATH"
+	JAVA_HOME="/usr/lib/jvm/java-8-openjdk-amd64/"
```
jwk404, line 40: This string is hard coded in lots of config.sh files. Would it be better to define a constant in common.sh and set JAVA_HOME that way? Or is sourcing that file a problem in some way?

jwk404 (approved):

packages/containerized-masking/config.sh
palash-gandhi: @sebroy mentioned that different packages should be able to define their own JAVA_HOMEs. I hear that masking wants to move to JDK 17 at some point so they might have to revert back to this.

PR #333, opened 2025-01-30
DLPX-93263 24.04 LTS: Set delphix-rust package version to rustc version

palash-gandhi:
See https://github.com/delphix/delphix-rust/pull/21

### DIFF
--- a/packages/delphix-rust/config.sh
+++ b/packages/delphix-rust/config.sh
@@ -20,7 +20,16 @@ DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/delphix-rust.git"
 
 function build() {
 	logmust mkdir -p "$WORKDIR/repo"
+
+	#
+	# Instead of relying on linux-pkg to assign a default version like 1.0.0, set the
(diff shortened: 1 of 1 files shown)


packages/delphix-rust/config.sh
```diff
@@ -20,5 +20,7 @@ DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/delphix-rust.git"
 
 function build() {
 	logmust mkdir -p "$WORKDIR/repo"
+	PACKAGE_VERSION="$(tr -d '\n' <"$WORKDIR/repo/RUSTC_VERSION")"
+	[[ -n "$PACKAGE_VERSION" ]] || die "Failed to retrieve package version"
```
sebroy, line 32: Can you add a comment explaining what we're doing here? I'm sure they next person to lay eyes on this will appreciate it.

sebroy (approved):

packages/delphix-rust/config.sh
palash-gandhi: Done.

prakashsurya (approved):

PR #335, opened 2025-02-14
DLPX-93401 LTS 24.04: Kernel builds require delphix-rust-src to be installed

palash-gandhi:
<details open>

Kernel builds continue to fail because `delphix-rust-src` is not installed
on the buildserver and apt tries to install `rust-src` and `rust-fmt` from
the Ubuntu repos and they conflict with our rust packages. This is why
`delphix-rust-src` is now built and should now be installed before building
the kernel.
</details>

<details open>

Install it before building the kernel.
</details>

<details open>

https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/os-upgrade/job/build-package/job/linux-kernel-generic/job/pre-push/57/console
</details>

### DIFF
--- a/default-package-config.sh
+++ b/default-package-config.sh
@@ -34,11 +34,18 @@ function merge_with_upstream() {
 #
 # The functions below are specific for the Linux kernel packages
 # and contain the majority of their common code.
+# This also installs the `delphix-rust` and `delphix-rust-src`
+# packages to satisfy the kernel's dependency on the rust toolchain
+# by ensuring that Delphix's version of the rust toolchain is
+# installed. Delphix's rust toolchain is supplied via virtual
+# packages and hence must be installed explicitly otherwise apt
+# installs the Ubuntu's version of the rust toolchain.
 #
 function kernel_prepare() {
 	logmust install_pkgs \
-		equivs \
+		"$DEPDIR"/delphix-rust/*.deb \
 		devscripts \
+		equivs \
 		kernel-wedge
 }
 
--- a/default-package-config.sh
+++ b/default-package-config.sh
@@ -34,8 +34,12 @@ function merge_with_upstream() {
 #
 # The functions below are specific for the Linux kernel packages
 # and contain the majority of their common code.
+#
 # Ensure that Delphix's version of the rust toolchain is
-# installed rather than Ubuntu's version.
+# installed rather than Ubuntu's version to satisfy the kernel's
+# dependency on the rust toolchain. Delphix's rust toolchain is
+# supplied via virtual packages and hence must be installed explicitly
+# otherwise apt installs the Ubuntu's version of the rust toolchain.
 #
 function kernel_prepare() {
 	logmust install_pkgs \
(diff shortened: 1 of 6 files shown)


default-package-config.sh
```diff
@@ -37,6 +37,7 @@ function merge_with_upstream() {
 #
 function kernel_prepare() {
 	logmust install_pkgs \
+		delphix-rust-src \
```
prakashsurya: let's also add delphix-rust here, even though it'll already be pre-installed.. just to be a bit more explicit and complete..

and would also be good to add a quick comment.. that we're pre-installing these, so they can satisfy the build requirements defined in the kernel's package definitions.. and we need to pre-install because APT isn't smart enough to resolve it auomatically due to virtual packages being in the mix..

prakashsurya (approved):

palash-gandhi:
This is not going to work since our internally built packages are not available to apt during a package build. appliance-build is the only place where internally built packages are made available to apt and hence, this is replaced by https://github.com/delphix/appliance-build/pull/793

packages/linux-kernel-aws/config.delphix.sh
```diff
@@ -22,6 +22,16 @@ UPSTREAM_GIT_URL="https://git.launchpad.net/~canonical-kernel/ubuntu/+source/lin
...
 UPSTREAM_GIT_BRANCH="none"
 
+#
+# This also installs the `delphix-rust` and `delphix-rust-src`
+# packages to satisfy the kernel's dependency on the rust toolchain
+# by ensuring that Delphix's version of the rust toolchain is
+# installed. Delphix's rust toolchain is supplied via virtual
+# packages and hence must be installed explicitly otherwise apt
+# installs the Ubuntu's version of the rust toolchain.
+#
```
mmaybee: This comment (and also in the other kernel versions) could be simplified a bit. I think it is good in the generic config. But here is could just say:
```
# Ensure that Delphix's version of the rust toolchain is
# installed rather than Ubuntu's version.
```

mmaybee (approved):

packages/linux-kernel-aws/config.delphix.sh
palash-gandhi: done

default-package-config.sh
```diff
@@ -34,11 +34,14 @@ function merge_with_upstream() {
 #
 # The functions below are specific for the Linux kernel packages
 # and contain the majority of their common code.
+# Ensure that Delphix's version of the rust toolchain is
+# installed rather than Ubuntu's version.
```
mmaybee: Heh, I think this instance of the comment should have continued to use the longer explanation:
```
# This also installs the `delphix-rust` and `delphix-rust-src`
# packages to satisfy the kernel's dependency on the rust toolchain
# by ensuring that Delphix's version of the rust toolchain is
# installed. Delphix's rust toolchain is supplied via virtual
# packages and hence must be installed explicitly otherwise apt
# installs the Ubuntu's version of the rust toolchain.
```

mmaybee (approved):

default-package-config.sh
palash-gandhi: Ah, that makes sense. I combined the 2 here and left the others as-is..

PR #339, opened 2025-03-11
DLPX-93639 LTS 24.04: Upgrade fails due to drgn conflict with upstream python3-drgn package

prakashsurya:
<details open>
Upgrade from 20.04 to 24.04 fails due to package name difference of our internal version of the "drgn" package, and ubuntu's package.
</details>

<details open>
The solution here, is to keep using our internal "drgn" package, rather than ubuntu's.
</details>

### DIFF
--- a/packages/drgn/config.sh
+++ b/packages/drgn/config.sh
@@ -1,0 +1,40 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019, 2025 Delphix
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# shellcheck disable=SC2034
+DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/drgn.git"
+
+UPSTREAM_GIT_URL="https://github.com/osandov/drgn.git"
+UPSTREAM_GIT_BRANCH="main"
+
+function prepare() {
+	#
+	# Strictly speaking libkdumpfile is not a hard prerequisite for
+	# drgn itself, but it is a hard requirement in our use-case as
+	# we do want to use drgn for kdump-compressed crash dumps.
+	#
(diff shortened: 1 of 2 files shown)


packages/drgn/config.sh
```diff
@@ -0,0 +1,40 @@
...
+UPSTREAM_GIT_URL="https://github.com/osandov/drgn.git"
+UPSTREAM_GIT_BRANCH="main"
+
+function prepare() {
+	#
+	# Strictly speaking libkdumpfile is not a hard prerequisite for
+	# drgn itself, but it is a hard requirement in our use-case as
+	# we do want to use drgn for kdump-compressed crash dumps.
+	#
+	logmust install_pkgs "$DEPDIR"/libkdumpfile/*.deb
```
palash-gandhi: What are we doing with libkdumpfile? In https://github.com/delphix/linux-pkg/pull/325, I also removed our internal fork of libkdumpfile in favor of libkdumpfile10 from Ubuntu repos

packages/drgn/config.sh
```diff
@@ -0,0 +1,40 @@
...
+	logmust install_build_deps_from_control_file
+}
+
+function build() {
+	logmust dpkg_buildpackage_default
+}
+
+function update_upstream() {
+	logmust update_upstream_from_git
```
palash-gandhi, line 33: Should the upstream of this package be Ubuntu's python3-drgn package instead of the upstream git repo?

packages/drgn/config.sh
prakashsurya: sorry, yes, I meant to remove this.. I presume the ubuntu package is named `libkdumpfile`? that's named in the drgn control file, so should pull in ubunutu's version via the line just below..

packages/drgn/config.sh
prakashsurya: no, I was thinking we'd keep using the upstream version.. up for debate, but I think that's relatively low maintenance, and keeping our debugging tools up to date with upstream seems useful..?

packages/drgn/config.sh
palash-gandhi: It's called [libkdumpfile10](https://packages.ubuntu.com/search?keywords=libkdumpfile) so we will need to modify drgn's control file I believe

packages/drgn/config.sh
palash-gandhi: We should ensure that the `update-package` workflow runs for `drgn`.

palash-gandhi (approved):

packages/drgn/config.sh
prakashsurya: OK, opened: https://github.com/delphix/drgn/pull/67

packages/drgn/config.sh
prakashsurya: OK, opened: https://github.com/delphix/devops-gate/pull/3703

PR #371, opened 2025-08-19
CP-12693 Sign kernel modules and image during kernel build (no shim) CP-12694 Sign ZFS modules after ZFS build (no shim) CP-12695 Sign connstat module after build (no shim)

tonynguien:
<details open>
<summary><h2> Background </h2></summary>
With Secure Boot, kernel image and modules needs to be be signed so that they can be verified on boot.
</details>

<details open>
The changes here updated the following package builds:

1. linux-kernel-* - includes CONFIG_MODULE_SIG_FORCE and CONFIG_MODULE_SIG_KEY such that kernel build uses our keys to sign modules. Also, signs vmlinuz after the build since kernel build does NOT sign kernel image.
2. ZFS - signs zfs.ko and spl.ko for all generated deb packages after the build 
3. connstat - signs connstat.ko for all generated deb packages after the build

</details>

<details open>

- Successful appliance builds for ESX, AWS, GCP, and Azure which imported created images and successfully booted VMs on those cloud platforms.
- On AWS, upgrade blackbox test also ran successfully.

Last successful build (08/23) - https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/11986/

Current build (08/25) - https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/11996/console

</details>

### DIFF
--- a/default-package-config.sh
+++ b/default-package-config.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright 2018, 2020 Delphix
+# Copyright 2018, 2025 Delphix
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -161,6 +161,44 @@ function kernel_build() {
 	logmust fakeroot debian/rules printenv "${debian_rules_args[@]}"
 
 	#
+	# Configure signing keys/certs before build
+	#
+	# CONFIG_MODULE_SIG_KEY is set to /var/tmp/sbkeys/signing_key.pem in
+	# resources/delphix_kernel_annotations
+	#
+	FLAVOUR=$platform
+	OBJ=debian/build/build-$FLAVOUR
+	CERTS=$OBJ/certs
+
+	# ensure the objdir + certs dir exist
+	mkdir -p "$CERTS"
+	download_keys
+
+	# provide the key the packaging expects INSIDE the objdir
+	# (symlink or copy)
+	logmust ln -sf "${SB_KEYS_DIR}/signing_key.pem" "$CERTS/signing_key.pem"
+	logmust chmod 600 "$CERTS/signing_key.pem"
+
+	# create the DER .x509 that sign-file needs from .crt)
+	logmust openssl x509 -in "${SB_KEYS_DIR}/db.crt" -outform DER -out "$CERTS/signing_key.x509"
+
+	# sanity checks
+	logmust test -s "$CERTS/signing_key.pem" || {
+		echo "missing signing_key.pem"
+		exit 1
+	}
+	logmust test -s "$CERTS/signing_key.x509" || {
+		echo "missing signing_key.x509"
+		exit 1
+	}
+	logmust openssl pkey -in "$CERTS/signing_key.pem" -noout >/dev/null || {
+		echo "key unreadable"
+		exit 1
+	}
+	SBSIGN_KEY="${SBSIGN_KEY:-$SB_KEYS_DIR/db.key}"
+	SBSIGN_CERT="${SBSIGN_CERT:-$SB_KEYS_DIR/db.crt}"
+
+	#
 	# The default value of the tool argument for mk-build-deps
 	# is the following:
 	# "apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends"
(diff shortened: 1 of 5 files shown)


packages/common.sh
```diff
@@ -0,0 +1,74 @@
+#
```
sebroy: I think this stuff probably belong in `lib/common.sh` which is already sourced by packages' `config.sh` scripts.

.github/scripts/verify-query-packages.sh
```diff
@@ -31,11 +32,11 @@ test ${#fields[@]} -eq 1
 test "${fields[0]}" == 'https://github.com/delphix/zfs.git'
 
 # Expect that "list all" outputs all directory names under packages/
-diff <(ls -1 packages | sort) <(./query-packages.sh list all 2>&1 | sort)
+diff <(ls -1 packages | grep -v common.sh | sort) <(./query-packages.sh list all 2>&1 | sort)
```
sebroy: Is this change because you added a file that is not a package directory to the `packages` directory? If so, then let's not do that. There's an existing pattern for where to place helper functions for various use cases in linux-pkg, let's use that?

packages/connstat/config.sh
```diff
@@ -16,6 +16,9 @@
 #
 # shellcheck disable=SC2034
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/../common.sh"
```
sebroy: `lib/common.sh` is already sourced, we shouldn't need another `common.sh`.

packages/connstat/config.sh
```diff
@@ -50,4 +53,32 @@ function build() {
...
+			connstat_mod=$(find "$temp_dir" -type f -name connstat.ko)
+			logmust kmodsign sha256 "$SBSIGN_KEY" "$SBSIGN_DER" "$connstat_mod" "$connstat_mod.signed"
+			logmust mv "$connstat_mod.signed" "$connstat_mod"
+			logmust modinfo -F signer "$connstat_mod"
+
+			# Repack the .deb"
+			update_md5sums "$temp_dir"
+			repack_deb "$connstat_pkg" "$temp_dir"
+		done
+	delete_keys
```
sebroy: A couple of thoughts on this:
1. This fundamental logic is identical to the logic for the `zfs` build, as such, this should be made into a common function that takes a list of deb files as input.
2. This common function can call `download_keys` and `delete_keys` on behalf of the caller.
3. Having this part of the linux-pkg build have inherent knowledge of the names of the packaged artifacts feels like a layering violation (a change to the packaged filenames will require changes in two places). Why not have the function simply find any kernel module (e.g. anything with a `.ko` extension) and sign it?

resources/delphix_kernel_annotations
```diff
@@ -2,6 +2,9 @@
 # FORMAT: 4
 # ARCH: amd64
 # FLAVOUR: amd64-aws amd64-azure amd64-generic amd64-gcp amd64-oracle
+#
+CONFIG_MODULE_SIG_KEY  		policy<{'amd64': '"/var/tmp/sbkeys/signing_key.pem"'}>
+CONFIG_MODULE_SIG_FORCE 			policy<{'amd64': 'y', 'arm64': 'y'}>
```
sebroy: We don't build for arm64, is that part required?

default-package-config.sh
```diff
@@ -23,6 +23,9 @@
 # be overriden.
 #
 
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+source "$SCRIPT_DIR/packages/common.sh"
+
```
sebroy: `lib/common.sh` should already have been sourced.

sebroy (changes requested):

packages/common.sh
tonynguien: Oh, I thought I looked. Yes, it makes sense to use the existing lib/common.sh Duh!

.github/scripts/verify-query-packages.sh
tonynguien: Yes, these changes would be unnecessary.

packages/connstat/config.sh
tonynguien: Good point. Yes, a function that takes one or list of package(s) and signs all .ko files makes sense. I'll do that.

resources/delphix_kernel_annotations
tonynguien: No, arm64 build is not required. It was an oversight.

default-package-config.sh
tonynguien: Agreed. Will update.

packages/connstat/config.sh
tonynguien: Agreed.

packages/zfs/config.sh
```diff
@@ -174,6 +174,10 @@ function build() {
 	done
 	logmust cd "$WORKDIR"
 	logmust mv "all-packages/"*.deb "artifacts/"
+
+	# Sign ZFS modules in all packages
+	zfs_modules=$(find "$WORKDIR/artifacts" -type f -name "zfs-modules-*.deb" ! -name "*-dbg*")
```
sebroy: uber nit: the things that find is returning are package names, not module names.

packages/connstat/config.sh
```diff
@@ -50,4 +50,8 @@ function build() {
 
 	logmust cd "$WORKDIR/repo"
 	logmust mv ./*deb "$WORKDIR/artifacts/"
+
+	# Sign the generated modules
+	connstat_modules=$(find "$WORKDIR/artifacts" -type f -name "connstat-module-*.deb" ! -name "*-dbg*")
```
sebroy: uber nit: these are connstat_packages, and not connstat_modules

sebroy (approved):

david-mendez1 (approved):

packages/zfs/config.sh
tonynguien: Yes, I'll update it.

packages/connstat/config.sh
tonynguien: Yes, I'll update.

PR #388, opened 2026-02-26
DLPX-96594 fix linux-kernel-generic upstream merge for 6.17

lyriclake:
<details open>

After updating to kernel 6.17, linux-kernel-generic was failing to build the `perf` tool with:
<code>
15:14:09  mmap.c: In function ‘perf_mmap__read’:
15:14:09  mmap.c:217:38: error: invalid use of undefined type ‘union perf_event’
15:14:09    217 |         if (diff >= (int)sizeof(event->header)) {
15:14:09        |                                      ^~
15:14:09  mmap.c:221:29: error: invalid use of undefined type ‘union perf_event’
15:14:09    221 |                 size = event->header.size;
15:14:09        |                             ^~
15:14:09  mmap.c:223:40: error: invalid use of undefined type ‘union perf_event’
15:14:09    223 |                 if (size < sizeof(event->header) || diff < (int)size)
15:14:09        |                                        ^~
15:14:09  make[5]: *** [/export/home/delphix/jenkins/workspace/linux-pkg/develop/build-package/linux-kernel-generic/pre-push/linux-pkg/packages/linux-kernel-generic/tmp/repo/debian/build/tools-perarch/tools/build/Makefile.build:86: /export/home/delphix/jenkins/workspace/linux-pkg/develop/build-package/linux-kernel-generic/pre-push/linux-pkg/packages/linux-kernel-generic/tmp/repo/debian/build/tools-perarch/tools/perf/libperf/mmap.o] Error 1
</code>
</details>

<details open>
I did not root-cause the perf compilation failure. However, in default-package-config.sh we have:

>        # do_tools_common=false
>        #   We do not need to build linux-tools-common package and we
>        #   install it directly from our package mirror.

We don't have any custom modifications to `perf`, and it is provided by linux-tools-common, so we don't need to be building it. This PR adds `do_tools_perf=false`.  
</details>

<details>
https://selfservice-jenkins.eng-tools-prd.aws.delphixcloud.com/job/appliance-build-orchestrator-pre-push/13331/

</details>

### DIFF
--- a/default-package-config.sh
+++ b/default-package-config.sh
@@ -122,6 +122,9 @@ function kernel_build() {
 	# do_tools_common=false
 	#   We do not need to build linux-tools-common package and we
 	#   install it directly from our package mirror.
+	# do_tools_perf=false
+	#   The perf tool is provided by the linux-tools-common package,
+	#   which we install directly from our package mirror.
 	# do_tools_host=false
 	#   We do not need to build linux-tools-host package.
 	# uefi_signed=false
@@ -136,6 +139,7 @@ function kernel_build() {
 	local debian_rules_args=(
 		"do_dbgsym_package=true"
 		"do_tools_common=false"
+		"do_tools_perf=false"
 		"do_tools_host=false"
 		"uefi_signed=false"
 		"do_zfs=false"


default-package-config.sh
```diff
@@ -122,6 +122,9 @@ function kernel_build() {
 	# do_tools_common=false
 	#   We do not need to build linux-tools-common package and we
 	#   install it directly from our package mirror.
+	# do_tools_perf=false
+	#   The perf tool is provided by the linux-tools-common package,
+	#   which we install it directly from our package mirror.
```
dbjwhs-perforce: nit: (sorry grammar police disregard if you want of course)

which we install it directly -> which we install directly

default-package-config.sh
```diff
@@ -122,6 +122,9 @@ function kernel_build() {
 	# do_tools_common=false
 	#   We do not need to build linux-tools-common package and we
 	#   install it directly from our package mirror.
+	# do_tools_perf=false
```
dbjwhs-perforce, line 125: nit: dunno if you want the comment items to be in same order as actual code?

dbjwhs-perforce (approved):
nit boy in the house.

default-package-config.sh
lyriclake: fixed, thank you!

sebroy:
I thought that while the perf tool's wrapper is delivered by that common package, there is a perf component that must be built with the kernel in order for the perf tool to work (otherwise that perf wrapper can't find the kernel artifacts to use in the filesystem).

I could be wrong, but maybe this doubt can be put to rest by testing that on an appliance with a 6.17 kernel built with this change, the perf binary works as expected.

lyriclake:
> I thought that while the perf tool's wrapper is delivered by that common package, there is a perf component that must be built with the kernel in order for the perf tool to work (otherwise that perf wrapper can't find the kernel artifacts to use in the filesystem).
> 
> I could be wrong, but maybe this doubt can be put to rest by testing that on an appliance with a 6.17 kernel built with this change, the perf binary works as expected.

I think you are correct here, thanks for pointing this out. Doing the test you suggest we see:
```
delphix@ip-10-110-246-204:~$ perf
WARNING: perf not found for kernel 6.17.0-1007
```
Luckily, in the course of testing this I also noticed that for whatever reason, linux-kernel-generic seems to be compiling successfully without this change now.. i.e., the error that prompted this PR no longer occurs. So I think it is safe to close this PR and merge the kernel changes without it.

PR #395, opened 2026-05-29
DLPX-97018 grub2 package doesn't need to be rebuilt

prakashsurya:
We're no longer updating grub from upstream, so, setting `DEFAULT_PACKAGE_GIT_URL` to `none` prevents some unnecessary sync work that is no longer relevant.


dbjwhs-perforce (approved):
LGTM

tperkins-perforce (approved):
lgtm

PR #407, opened 2026-08-05
CP-13581 linux-pkg: build windows-connector on-demand, caching result in Artifactory third-party-local instead of rebuilding every run

vimleshmishra:
<details open>

The `windows-connector` package is rebuilt from scratch on every `linux-pkg` pipeline run, even when its version (`project.ext.connectorVersion` in `appliance/host/windows/build.gradle`) hasn't changed since the last successful build. This is unnecessarily slow, and this rebuild-every-time flow was also masking the underlying dependency-drift bug (see [dlpx-app-gate#4648](https://github.com/delphix/dlpx-app-gate/pull/4648)) — the failure only surfaces when a buildserver's live apt mirror happens to have moved out of sync with the previously-pinned Artifactory dependency zips.
</details>

<details open>

Change `config.sh`'s `build()` so the package is only rebuilt when its version changes:
- If `windows-connector_<version>_all.deb` already exists at `s3://release-de-images/windows-connector-cache/<version>/`, download it directly instead of rebuilding.
- Otherwise, build it (`gradlew build` + `createDebPackage`) and upload the result to that same path so future runs of the same version can reuse it.

Chose `release-de-images` because it already exists specifically for long-lived build/release artifacts and has **no object-expiration policy** (unlike `snapshot-de-images`, which expires objects after 30 days — too short for a cache that may go untouched for months between connector version bumps). This uses the `aws-linux-build-publisher` AWS credential that's already bound unconditionally for every `linux-pkg` package build, so **no new credential or permission grant is needed anywhere**.
</details>

<details open>

Tested end-to-end on **real production Jenkins**, both code paths:
- Build [#934](https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/develop/job/build-package/job/windows-connector/job/post-push/934/) — version `2.3.0.0` not cached: `gradlew build` → `BUILD SUCCESSFUL`; `gradlew createDebPackage` → `BUILD SUCCESSFUL`; `.deb` uploaded cleanly to `s3://release-de-images/windows-connector-cache/2.3.0.0/`
- Build [#935](https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/develop/job/build-package/job/windows-connector/job/post-push/935/) — version `2.3.0.0` already cached: build skipped entirely, `.deb` downloaded from S3, job finished in ~5 min instead of ~10
</details>

<details open>
<summary><h2> Notes to Reviewers </h2></summary>

No devops-gate changes are required for this PR — the AWS credential used here is already available to every package build.

Tracked under epic [CP-13579](https://perforce.atlassian.net/browse/CP-13579), alongside [CP-13582](https://perforce.atlassian.net/browse/CP-13582) (app-gate).
</details>


packages/windows-connector/config.sh
```diff
@@ -18,18 +18,58 @@
...
+
+		if [[ -z "$BLACKBOX_USERNAME" || -z "$BLACKBOX_PASSWORD" ]]; then
+			echo "WARNING: BLACKBOX_USERNAME/BLACKBOX_PASSWORD not set - skipping Artifactory upload." \
+				"Upload $WORKDIR/artifacts/$deb_name to $artifactory_dir/ manually so future builds can reuse it."
+		else
+			echo "Uploading $deb_name to $artifactory_dir/"
+			local http_code
+			http_code=$(curl -s -o /dev/stderr -w "%{http_code}" \
+				-u "${BLACKBOX_USERNAME}:${BLACKBOX_PASSWORD}" \
+				-T "$WORKDIR/artifacts/$deb_name" "$artifactory_dir/$deb_name")
```
prakashsurya: I'm confused.. why are we putting this in artifactory?

we already support storing in S3, automatically.. and S3 works properly for all our various build cases (pre-push, post-push, hotfixes, etc..)..

why is S3 insufficient?

packages/windows-connector/config.sh
```diff
@@ -18,18 +18,58 @@
...
+	local version
+	version=$(grep "project.ext.connectorVersion" "$INSTALLER_DIR/build.gradle" |
+		sed -E "s/.*'([^']+)'.*/\1/")
+	local artifactory_dir="http://artifactory.delphix.com/artifactory/third-party-local/windows-connector/$version"
+	local deb_name="windows-connector_${version}_all.deb"
+
+	if wget --spider -q "$artifactory_dir/$deb_name"; then
+		echo "windows-connector $version already built; downloading pre-built installer from Artifactory"
+		logmust cd "$WORKDIR/artifacts"
+		logmust fetch_file_from_artifactory "$artifactory_dir/$deb_name"
```
prakashsurya: this doesn't make any sense to me.. why are we pulling an artifact from artifactory here, rather than doing the build?

is this because the build job will get triggered any time we change the app-gate, and that might include changes that don't include any changes to windows-connector? so we're optimizing for this case, to avoiding a build, when a build is unnecessary?

if so, why do we need such optimization? why not just do the build regardless? is it really too costly to do the build, that we need to add this complexity to avoid it?

prakashsurya (changes requested):

packages/windows-connector/config.sh
```diff
@@ -18,18 +18,58 @@
 
 DEFAULT_PACKAGE_GIT_URL="https://github.com/delphix/dlpx-app-gate.git"
 SKIP_COPYRIGHTS_CHECK=true
+PACKAGE_GIT_BRANCH="dlpx/pr/vimleshmishra/716aac3d-399f-43c7-9b80-c373ab6f7218"
```
prakashsurya: I'm assuming this is temporary, but making a note to ensure it doesn't land accidentally.

packages/windows-connector/config.sh
vimleshmishra: Good catch — this is already resolved. We pivoted this PR to S3 (`release-de-images`) for exactly the reason you're describing: it already supports our various build cases and has no object-expiration policy, so there's no need for a new Artifactory permission grant. The current diff no longer touches Artifactory at all.

Tested end-to-end on real production Jenkins:
- Build [#934](https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/develop/job/build-package/job/windows-connector/job/post-push/934/) — version not cached, built from source, uploaded to `s3://release-de-images/windows-connector-cache/2.3.0.0/`
- Build [#935](https://ops-jenkins.eng-tools-prd.aws.delphixcloud.com/job/linux-pkg/job/develop/job/build-package/job/windows-connector/job/post-push/935/) — version already cached, build skipped entirely, downloaded from S3

Will push the updated diff shortly.

packages/windows-connector/config.sh
vimleshmishra: Fair question, and worth answering honestly rather than defensively. Once #4648 (the app-gate fix) lands, the drift-failure root cause is gone regardless of this caching layer — mono/mingw/nsis install live from apt every time either way. So the justification for this PR is narrower than how it was originally framed: it's purely a build-time optimization, not a correctness fix.

In testing, a full `gradlew build` + `createDebPackage` takes ~7-8 minutes (build #934: 5m15s + 2m23s), which gets skipped entirely whenever windows-connector's own version hasn't changed (build #935 finished in ~5 min instead of ~10, since it skipped straight to downloading the cached `.deb`). windows-connector currently rebuilds on every linux-pkg/app-gate-triggered pipeline run regardless of whether anything in it actually changed, so that adds up across builds that have nothing to do with windows-connector at all.

That said, I don't want to oversell it — if ~7-8 min/build doesn't seem worth the added complexity here, I'm open to reconsidering scope (e.g. dropping this PR and just shipping #4648 on its own). Let me know which way you'd lean.

vimleshmishra: One more angle worth adding, since it strengthens the case beyond just raw minutes: the "why not just always build" question also runs into how these two things actually change at very different rates. windows-connector itself changes rarely — the version only bumps when there's an actual functional change to what ships to the Windows host. The apt/mono/mingw/nsis toolchain packages used to build it, on the other hand, get updated on their own schedule, independent of this installer, and that churn is purely toolchain-level — it never changes anything about the shipped Windows binary/installer itself.

Rebuilding on every pipeline run means re-running (and implicitly re-validating) this installer against toolchain drift that has zero effect on it, on every single build — including builds that don't touch windows-connector at all. That's not just ~7-8 min of compute repeated for no reason; it's also re-spending test/QA effort re-certifying an artifact that hasn't functionally changed, purely because something unrelated (an apt mirror snapshot) moved underneath it. Caching by `connectorVersion` lines the rebuild/re-test cadence up with the artifact's actual rate of change instead of the toolchain's.

packages/windows-connector/config.sh
prakashsurya: Why do you need to use `release-de-images` ? I still don't get it.. the last built package, of any package in linux-pkg, is preserved indefinitely, regardless of the bucket's expiration time..

so, why explicitly use `release-de-images`, and not just do w/e the default is for linux-pkg.. which supports pre-push, post-push, hotfixes, etc..

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,47 @@ function prepare() {
...
 
+#
+# windows-connector is only rebuilt when its version (project.ext.connectorVersion
+# in appliance/host/windows/build.gradle) is bumped. If that version has already
+# been built and uploaded to S3, we just download the existing installer instead
+# of rebuilding. Otherwise we build it and upload the result to S3 so future builds
+# of the same version can reuse it.
+#
+# release-de-images (unlike snapshot-de-images, which expires objects after 30
+# days) has no expiration policy, so the cached installer persists indefinitely -
```
prakashsurya: the last built package does not expire.. if it did, all of our builds would break.. I really don't think we have a problem here..

can you please explain why the default expiration of our S3 buckets is a problem, and why you think this specific linux-pkg package needs something customized, w.r.t. all other linux-pkg packages?

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,47 @@ function prepare() {
...
+	local version
+	version=$(grep "project.ext.connectorVersion" "$INSTALLER_DIR/build.gradle" |
+		sed -E "s/.*'([^']+)'.*/\1/")
+	local s3_dir="s3://release-de-images/windows-connector-cache/$version"
+	local deb_name="windows-connector_${version}_all.deb"
+
+	if aws s3 ls "$s3_dir/$deb_name" >/dev/null 2>&1; then
+		echo "windows-connector $version already built; downloading pre-built installer from S3"
+		logmust cd "$WORKDIR/artifacts"
+		logmust aws s3 cp "$s3_dir/$deb_name" .
```
prakashsurya, line 56: is performing a 8min build really that important to optimize out? why not just do the build, vs. add all this complexity w.r.t. using a "cache"..?

perhaps I'm misguided, but I just don't see this as necessary.. and the additional complexity, and inconsistency with all other linux-pkg packages, and more of a burden than performing the build itself..

packages/windows-connector/config.sh
prakashsurya: > it's also re-spending test/QA effort re-certifying an artifact that hasn't functionally changed

what do you mean by this? I don't buy this as a good reason to need a cache, but I'm open to learning more; perhaps I'm wrong.

vimleshmishra: > I don't buy this as a good reason to need a cache, but I'm open to learning more; perhaps I'm wrong.

Understood, and to be direct: this is the actual, primary reason I want the cache — not a secondary justification stacked on top of the build-time argument. The build-time savings are a nice side effect, not the driver.

nsis/mono/mingw-w64 aren't part of the connector's own source — they're the open-source build toolchain that replaced a licensed Visual Studio toolchain we used to build this installer with. They're inputs to the build, not something we own or validate the correctness of ourselves. Right now, every rebuild bakes in whatever versions of those packages happen to be live in Artifactory at that moment — regardless of whether our own connector source changed at all. So a rebuild triggered by something totally unrelated (this PR started from exactly that: a DLPX-97587 hotfix that never touches windows-connector) still produces a technically-different binary, assembled from a different toolchain combination, without anyone deciding to test and ship that new combination. We happen to still run the install test suite against it, so it's not shipping untested — but that's an accidental side effect of "we always rebuild," not a deliberate choice to re-certify a new toolchain mix.

What we want instead is to make that a deliberate choice: certify a given windows-connector version once, against one specific toolchain snapshot, and then guarantee that exact certified artifact — not some new recombination of tool versions — is what ships until we actually change our own connector source. That matters more here than for most packages because this installer isn't tied to the engine upgrade path — customers manually uninstall/reinstall it. Bit-for-bit consistency over time carries more weight for something this manually handled and infrequently touched than it does for something redeployed with every release.

Related: we're also moving away from pinning these binaries to a fixed version number at all. Staying pinned against a live, drifting apt mirror is exactly what let inconsistency creep in silently — the connector could end up behaving differently depending on which drifted snapshot got resolved into a given build. That's the kind of unnecessary inconsistency we don't want showing up as different connector behavior across customer environments.

There's a broader version of this discussion happening in the [Slack thread](https://perforce.slack.com/archives/C07H4GRPXUN/p1785940841105469) (build-time framing, the `dev-de-images`/"latest" alternative, and a couple other points) — worth reading together with this so we're not answering the same question two different ways in two places.

packages/windows-connector/config.sh
vimleshmishra: Same answer as the other thread — it's not really about the 8 minutes, that's a side effect, not the driver. The reason is toolchain-drift/certification consistency: mono/mingw/nsis are an external, independently-drifting build toolchain (they replaced a formerly-licensed Visual Studio toolchain for this installer), not our own source. Rebuilding on every trigger — including ones that touch nothing in windows-connector — means the resulting binary can silently vary based on whatever toolchain snapshot Artifactory happens to resolve at that moment, without anyone deciding to re-certify that combination. This installer also isn't tied to the engine upgrade path — customers manually uninstall/reinstall it — so that kind of untracked variability matters more here than it would for a package that gets redeployed with every release.

On the inconsistency with other linux-pkg packages: that's intentional, not an oversight. windows-connector is the one package in linux-pkg built from this kind of external, drifting toolchain dependency — the rest are native builds without an equivalent risk, so they don't need the same handling.

Full context (including the `dev-de-images`/"latest" alternative and a couple other angles) is in the [Slack thread](https://perforce.slack.com/archives/C07H4GRPXUN/p1785940841105469) — worth reading alongside this one.

packages/windows-connector/config.sh
vimleshmishra: Fair point, and I'm not going to argue persistence/expiration — that was never actually the driving reason, and it's already been dropped from the code comment as part of the latest push (see the updated `config.sh`). Sorry for the confusion that framing caused.

The actual reason for the cache is toolchain-drift/certification consistency: mono/mingw/nsis are an external build toolchain (formerly a licensed Visual Studio toolchain) that can drift independently of our own connector source. Rebuilding on every trigger — even ones unrelated to windows-connector — means the shipped binary can silently vary based on whatever toolchain snapshot happens to resolve at that moment, without anyone deciding to re-certify that combination. Caching by `connectorVersion` guarantees the exact certified artifact ships until our own source actually changes, which matters more here since this installer isn't tied to the engine upgrade path — customers manually uninstall/reinstall it.

On why this package specifically vs. every other linux-pkg package: windows-connector is the one package built from this kind of external, drifting toolchain — the rest are native builds without that dependency, so they don't need the same handling.

Same [Slack thread](https://perforce.slack.com/archives/C07H4GRPXUN/p1785940841105469) has the fuller discussion if useful.

packages/windows-connector/config.sh
vimleshmishra: Not landing as originally shown — this design's since been fully replaced (see latest diff). No more release-de-images/Artifactory at all.

packages/windows-connector/config.sh
vimleshmishra: Agreed, and that's exactly what the latest diff does — dropped `release-de-images` entirely, now reuses the same default latest-artifact lookup (`get_package_dependency_s3_url`) every other linux-pkg package already uses.

packages/windows-connector/config.sh
vimleshmishra: Superseded — this design has since changed entirely (see updated diff/description): dropped Artifactory/release-de-images, now reuses the existing default `latest` lookup every linux-pkg package already uses.

packages/windows-connector/config.sh
vimleshmishra: Agreed, and moot now — the expiration/persistence argument was dropped from the design. Current diff reuses the existing `latest` mechanism, which already persists indefinitely via the standing `s3-artifact-refresh` job.

packages/windows-connector/config.sh
vimleshmishra: Also superseded — no longer about build time. Latest diff's rationale is toolchain-drift/certification consistency (see updated PR description), and it no longer needs any new bucket/credential to do it.

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,82 @@ function prepare() {
...
+	if [[ -n "$latest_s3_url" ]]; then
+		local bucket="${latest_s3_url#s3://}"
+		bucket="${bucket%%/*}"
+		local key="${latest_s3_url#s3://"$bucket"/}"
+
+		if aws s3api head-object --bucket "$bucket" --key "$key/$deb_name" >/dev/null 2>&1; then
+			echo "windows-connector $version already built (latest); reusing it instead of rebuilding"
+			logmust cd "$WORKDIR/artifacts"
+			logmust aws s3 cp --only-show-errors "$latest_s3_url/$deb_name" .
+			reused=true
```
sebroy: This is not needed, and wastes s3 object space for nothing. The previously built package is already referenced by the "latest" combine-packages object, so there is no reason to have to copy that and call combine-packages again. IMO we should do nothing at all in this case. There is an explicit Jenkins build status to represent that something wasn't built ("NOT_BUILT"). That feels appropriate here. We should exit with some code that results in the build-package job ending as "NOT_BUILT".

sebroy (changes requested):

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,82 @@ function prepare() {
...
+		# syncs $WORKDIR/artifacts to S3 and updates "latest" for us, exactly
+		# like every other linux-pkg package.
+		#
+	fi
+
+	if [[ "$reused" == true ]]; then
+		echo "reused" >"$WORKDIR/artifacts/BUILD_STATUS"
+	else
+		echo "built_fresh" >"$WORKDIR/artifacts/BUILD_STATUS"
+	fi
```
prakashsurya: why does this matter? this essentially equates to it always being "reused", due to the cadence of app-gate pushes to connect version bumps.. regardless, what consumes this, and why?

feels overly complicated for no good reason, IMO..

packages/windows-connector/config.sh
prakashsurya: I agree.

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,82 @@ function prepare() {
...
+		logmust cd "$CONNECTOR_DIR"
+		logmust sudo ../../gradlew build
+		logmust cd "$INSTALLER_DIR"
+		logmust sudo ../../gradlew createDebPackage
+		logmust cp ./build/distributions/*.deb "$WORKDIR/artifacts/$deb_name"
+		#
+		# No explicit upload here -- the shared post-push Publish stage already
+		# syncs $WORKDIR/artifacts to S3 and updates "latest" for us, exactly
+		# like every other linux-pkg package.
+		#
```
prakashsurya: unnecessary comment, this is core behavior provided by the linux-pkg infra, and relied on by every package.. don't think it needs a package specific comment.

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,82 @@ function prepare() {
...
-	logmust cd "$CONNECTOR_DIR"
-	logmust sudo ../../gradlew build
-	logmust cd "$INSTALLER_DIR"
-	logmust sudo ../../gradlew createDebPackage
-	logmust sudo mv ./build/distributions/*deb "$WORKDIR/artifacts/"
+
+	local version
+	version=$(grep "project.ext.connectorVersion" "$INSTALLER_DIR/build.gradle" |
+		sed -E "s/.*'([^']+)'.*/\1/")
+	local deb_name="windows-connector_${version}_all.deb"
```
prakashsurya, line 41: this assumes a rebuild will produce an artifact of this specific filename.. without actually controlling it here.. IMO, that coupling is fragile, as the filename can change in the app-gate repo, and this repo would then not behave properly (e.g. it'd start to rebuild when we don't want it to)..

prakashsurya (commented):
I agree with seb's latest comment..

IMO, if we really want to keep the post-push build trigger on the app-gate git repository, then I think this is overly complicated.. and really, we want a way to denote the build as "skipped".. instead of the S3 file manipulation.. and do nothing when it's skipped, as the prior artifact will already be preserved "forever"..

With that said, since we want to tightly control when we do rebuilds of the connector, I might suggest an even more simple approach.. and disable the post-push trigger entirely, and instead rely on folks to manually trigger new builds.. then, we can carefully control when we rebuild it, and tie that into w/e "recertification" process we currently have (correct me if this doesn't make sense, I'm unfamiliar with the current certification process that's mentioned in the PR description).. and the build script can just be simple like it is today, and always do a build..

To me, this feels awfully complicated for what essentially equates to "do nothing, if connector version hasn't changed".. there's no need to copy the S3 artifacts around..

packages/windows-connector/config.sh
vimleshmishra: Agreed, done — dropped the copy-forward entirely. On a cache hit, `build()` now just exits with a sentinel that the pipeline maps to a real Jenkins `NOT_BUILT` result. No download, no republish, no wasted space.

packages/windows-connector/config.sh
vimleshmishra: Agreed, removed — this was the `BUILD_STATUS` file, which is gone now that the build genuinely exits `NOT_BUILT` on a skip. Jenkins' own build result carries that signal natively, so there's nothing left to write or consume.

packages/windows-connector/config.sh
vimleshmishra: Removed.

packages/windows-connector/config.sh
vimleshmishra: Fair concern, and there's actually a cleaner fix than defending this: the real problem is that `build.gradle`'s `createDebPackage` names its output using the current timestamp, not `connectorVersion` — so it's not version-derived at all today. Fixing that in app-gate (naming the artifact from `connectorVersion` directly) removes the need for any renaming here entirely. That's a separate, app-gate-side PR. This file is already updated to assume that fix — the rebuild path now does a plain `cp ./build/distributions/*.deb "$WORKDIR/artifacts/"` with no explicit rename, so once that lands, this just works with zero further changes here.

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,51 @@ function prepare() {
...
+		local key="${latest_s3_url#s3://"$bucket"/}"
+
+		if aws s3api head-object --bucket "$bucket" --key "$key/$deb_name" >/dev/null 2>&1; then
+			echo "windows-connector $version already built (latest); nothing to do"
+			#
+			# Marker file, not a special exit code: the Jenkins pipeline checks for
+			# this file's existence to set a NOT_BUILT result instead of relying on
+			# a distinguished exit code from this stage.
+			#
+			touch "$WORKDIR/artifacts/.build_skipped"
```
sebroy, line 63: This should be done through an official contract/interface rather than a windows-connector one-off. The other issue here is that by the time `build()` runs, we've already done `prepare()` and wasted time. My suggestion is to add a new `check_rebuild_needed` optional hook that only this package implements. Then, the logic of how that is handled can live in the generic buildpkg.sh script.

packages/windows-connector/config.sh
```diff
@@ -24,12 +24,51 @@ function prepare() {
...
+# "latest" artifact lookup every package already uses for its build dependencies;
+# fails soft (falls through to a real build) if nothing is found there.
+#
 function build() {
 	CONNECTOR_DIR="${WORKDIR}/repo/appliance/server/connector"
 	INSTALLER_DIR="${WORKDIR}/repo/appliance/host/windows"
+
+	local version
+	version=$(grep "project.ext.connectorDebVersion" "$INSTALLER_DIR/build.gradle" |
+		sed -E "s/.*'([^']+)'.*/\1/")
```
sebroy, line 40: need error handling

sebroy (changes requested):

PR #410, opened 2026-08-05
TOOL-30737 unblock manual package builds against the upstream Ubuntu archive

prakashsurya:
Fixes **TOOL-30737** and **TOOL-30738**, one commit each. The title carries only
the first because the pre-push Jira check accepts a single issue id.

Both were found while doing a manual
`resolute` (Ubuntu 26.04) package build to exercise the build container from
#404 ahead of the `os-upgrade` mirror sync.

### TOOL-30737 — secondary mirror blocks building against upstream Ubuntu

`setup.sh` writes two kinds of apt source: the primary mirror (a snapshot of the
Ubuntu archive) and the secondary mirror (the PPAs). `DELPHIX_PACKAGE_MIRROR_MAIN`
and `..._SECONDARY` already let a caller repoint either.

The primary has an upstream equivalent; the secondary does not. Pointing the
primary at `archive.ubuntu.com` is enough to build a suite the Delphix mirror has
never carried — which is what the start of an LTS upgrade looks like, since the
new suite exists upstream months before its mirror sync publishes. But the
secondary's line is written unconditionally with nothing to point it at:

```
E: The repository '.../ppas resolute Release' does not have a Release file
```

That fails `setup.sh`, so *no* package builds — including the many whose build
dependencies come entirely from the Ubuntu archive. `..._SECONDARY=none` now
omits the line and its keyring. Leaving the variable unset behaves exactly as
before.

### TOOL-30738 — `buildpkg.sh -l` has never worked

```
cp -r "$WORKDIR/../../$dep/tmp/artifacts/ $dep/"
```

Both operands are inside one set of quotes, so `cp` gets a single argument and
exits with `missing destination file operand`. Broken since `7fe9cd2` (2022). It
looked fine because the loop only runs for packages declaring
`PACKAGE_DEPENDENCIES` — every other package returns early, and `-l` is a no-op
for them by definition. So it appeared to work everywhere except the one case it
exists for.

Unrelated to the container work; the same testing just surfaced it.

## Testing

Provisioned a fresh `dlpx-internal-buildserver-develop` VM (`COMPUTE_XLARGE`) —
a **noble** host — and built for `resolute` against
`us-west-2.ec2.archive.ubuntu.com`, with `DELPHIX_PACKAGE_MIRROR_SECONDARY=none`.
No `DISABLE_SYSTEM_CHECK` needed, since it is a real buildserver image.

A representative subset, chosen to cover distinct build shapes:

| Package | Result | Shape exercised |
|---|---|---|
| `libkdumpfile` | PASS 122s | autotools C, `mk-build-deps` from `debian/control` |
| `cloud-init` | PASS 209s | Ubuntu-forked source package, python/debhelper |
| `crypt-blowfish` | PASS 66s | plain `make`, private repo via `FETCH_GIT_TOKEN` |
| `docker-python-image` | PASS 123s | `PACKAGE_NEEDS_DOCKER` — socket bind + `--group-add` |
| `delphix-platform` | PASS 81s | in-house, `sudo -E make build-deps` / `make packages` |
| `gdb-python` | FAIL 109s | cross-package `PACKAGE_DEPENDENCIES` via `-l` |

The artifacts are genuinely resolute-linked rather than incidentally produced:
`libkdumpfile`'s generated `Depends` pin `libbinutils (>= 2.46)` and
`libsnappy1v5 (>= 1.2.2)`, which are resolute versions — noble ships binutils
2.42 and snappy 1.1.10 — so `dpkg-shlibdeps` resolved against the container's
rootfs, not the host's. `docker-python-image` really did reach the host daemon
(`--group-add 990`, a live `docker pull` from `registry.delphix.com`, 142 MB
deb).

`gdb-python`'s failure is **not** a defect in either change and is deliberately
left alone here: `python3-future` does not exist in resolute, so its
`debian/control` needs updating on the `os-upgrade` branch. It is Phase 2 work
for CP-13555, and exactly the class of finding this exercise is meant to
surface — the build got all the way to dependency resolution against a real
resolute archive before hitting it. The `-l` fix is load-bearing for reaching
that point: it is what let `gdb-python` consume the resolute `libkdumpfile` built
122s earlier instead of the noble one in S3.

`make check` (shellcheck + shfmt) clean before and after.


prakashsurya:
## noble regression check

Verified the ordinary **noble** path is unaffected, on the same buildserver, from
this branch. Deliberately **no** mirror overrides, so both URLs resolve the real
Delphix `develop` snapshot (`.../develop/3ce55a6a/{ubuntu,ppas}`) — i.e. exactly
what Jenkins does.

The secondary mirror still behaves as before when the variable is unset:

```
Running: sudo install -D ... resources/delphix-secondary-mirror.key /etc/apt/keyrings/delphix-secondary-mirror.gpg
Get:5  .../develop/3ce55a6a/ppas noble InRelease [9,356 B]
Get:51 .../develop/3ce55a6a/ppas noble/main amd64 Packages [534 kB]
Get:52 .../develop/3ce55a6a/ppas noble/stable amd64 Packages [62.9 kB]
```

The keyring is installed and the source line written and fetched — the `none`
branch is not taken.

| Package | noble | resolute |
|---|---|---|
| `libkdumpfile` | PASS 72s | PASS 122s |
| `gdb-python` (`-l`) | **PASS 1096s** | FAIL — `python3-future` absent |
| `cloud-init` | PASS 187s | PASS 209s |
| `delphix-platform` | PASS 63s | PASS 81s |

Two things this pins down:

**`gdb-python` passing on noble** is the end-to-end proof of the `-l` fix — the
two-argument `cp` runs, the `libkdumpfile` built 72s earlier in the same run
installs as the dependency, and a deb comes out. It also isolates the resolute
failure to `python3-future` alone, rather than anything about `-l` or the
container.

**Same source, same commit, different suite** produces correctly different
`shlibdeps`: noble pins `libbinutils (>= 2.42)` / `libsnappy1v5 (>= 1.1.10)`,
resolute pins `2.46` / `1.2.2`.

### Unrelated fragility worth a follow-up

The first noble `setup.sh` failed at `debootstrap` with
`E: Couldn't download packages: libassuan0 libatomic1 libattr1`. All three were
reachable (HTTP 200) from that host moments later and the retry succeeded, so it
was transient. `container_build_base_image()` calls `debootstrap` with no retry,
so a momentary mirror hiccup fails the whole build — the same class of problem
the mirror-sync job got `run_subprocess` retries for under TOOL-30598. Not
touched here; pre-existing in #404.

dbjwhs-perforce (approved):
LGTM

lyriclake (approved):

